PR #9, opened 2016-11-18
Change mkChar to mkCharLen for usage with single Chars instead of str…

alexreeder:
…ing. This does fix Burke-Lauenroth-Lab/Rsoilwat#22


dschlaep:
Compiles and executes the testing/ on my macOS.

PR #11, opened 2017-01-02
SOILWAT clean-up 

charlieduso:
This branch was created to fix some of the statements that produce warnings during compilation of SOILWAT (this excludes RSoilwat). As of my testing on two Ubuntu platforms, all of the compilation warnings were suppressed.

_Note: I had to void-cast some input parameters in the function "adjust_Tsoil_by_freezing_and_thawing()" as they were never used._

_Note: I also updated the version of SOILWAT within compilation and renamed the "files_v31.in" file to "files.in". When executing SOILWAT, the program would look for "files.in" and be unable to find it because of the previous convention._

 **I encourage someone else to test the compilation of this version of SOILWAT on their computer and various OS's if they have the option to.**

_I will be looking to fix compilation warnings with RSoilwat at a later time._


charlieduso:
Need to check and verify two more changes. This is an oversight by me.

Issue #19, opened 2017-06-26
Add unit tests


dschlaep:

dschlaep:
We decided to use [GoogleTest](https://github.com/google/googletest) as testing framework.

N1ckP3rsl3y:
GoogleTests were implemented, however this issue is ongoing/long-term as code is updated.

PR #70, opened 2017-09-29
Feature add gtest

### LINKED ISSUE #67: Set up GoogleTest as testing framework

dschlaep:
Unit testing framework
- close #67

Remaining work
- include tests on CIs (see #69)
- add SOILWAT2 relevant tests (see #19)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@
 # Google unit test output
 gtest-*.o
 libgtest.a
+sw_test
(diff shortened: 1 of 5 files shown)


dschlaep:
@kpalmqui could you please make sure that the changes made on this branch do not interfere with STEPWAT2? I don't believe that they would affect STEPWAT2 because all changes are restricted to subfolders 'doc/', 'test/', and 'googletest/' and their new make targets which are only activated through their own separate functionality. I just wanted to be sure that none of this causes any unwanted trouble, before I merge into master. Thanks!

PR #80, opened 2017-11-02
Beta random generator was added

2hua:
Beta random generator was added for the submodule called by StepWat. Maybe you could do some reviews about anything not reasonable. 
@dschlaep @kpalmqui


dschlaep:
- Consider following the naming scheme used for the other already implemented random number generators `RandUni`, `RandNorm`, ...
- Format documentation as [`Doxygen` comments](https://www.stack.nl/~dimitri/doxygen/manual/docblocks.html)
- I am unfamiliar with the term `random deviate` for the outcome of a beta distributed process; I would call this a `random variate` (e.g., https://en.wikipedia.org/wiki/Random_variate) -- I am only familiar with `random deviate` in the sense of a standard-normally distributed process (e.g., https://en.wikipedia.org/wiki/Standard_normal_deviate) when the meaning of  'deviation' makes directly sense.
- Update `Doxygen` documentation (see Readme)
- Did you develop the code from scratch? Or did you copy-paste from some external source? If the latter, then provide appropriate reference(s).
- Ideally, you would provide some unit tests (but I can understand that we would postpone this for now).
- You will have to update with `master` before you can merge into `master`.

dschlaep:
commit d243d0c61c9e9449de084cd3cd9c9270480df4d4 closes #69

dschlaep:
Caitlin and I have been addressing all of my comments above except one:
- Did you develop the code from scratch? Or did you copy-paste from some external source? If the latter, then provide appropriate reference(s) and license statements.

PR #102, opened 2017-12-07
Bugfix 88 rSOILWAT2

dschlaep:
- `LogError` writes to R console only if `fp` file pointer is not null
- this is a mechanism to implement quiet runs


dschlaep:
- `LogError` writes to R console only if `fp` file pointer is not null
- this is a mechanism to implement quiet runs

Issue #30, opened 2017-07-21
Reduce `#if defined(RSOILWAT)` directives to minimum


dschlaep:

dschlaep:
There are currently about 200 such directives in the code.

PR #126, opened 2018-01-07
Bugfix 66 rSOILWAT2

dschlaep:
- Move rSOILWAT2 code to rSOILWAT2 repository

- define global variables for number and names of output columns (and functions to set them)

- removed inefficiencies in output code (some loops, e.g., over time periods, repeated themselves)


dschlaep:
This will close #30.

Issue #31, opened 2017-07-21
Reduce #if defined(STEPWAT) directives to minimum


dschlaep:

dschlaep:
There are currently about 60 such directives in the master branch and about 65 in the 'soilwat_save_output' branch.

PR #194, opened 2018-06-15
Bugfix 177

dschlaep:
Fix and re-organize output functionality across `SOILWAT2-standalone`, `rSOILWAT2`, and `STEPWAT2`


dschlaep:
close #30, close #31

PR #198, opened 2018-06-21
Feature debug

dschlaep:
easier debugging with specific make targets


dschlaep:
@CaitlinA @nip5 This could be helpful to address issue #191.

Issue #208, opened 2018-07-10
clang: warning: treating 'c' input as 'c++' when in C++ mode, this behavior is deprecated


dschlaep:
Replicate that warning by using clang++ as the compiler, i.e.,
```
CXX=clang++ make cleaner test test_run


clang: warning: treating 'c' input as 'c++' when in C++ mode, this behavior is deprecated [-Wdeprecated]
```

(see, e.g., https://github.com/DrylandEcology/SOILWAT2/issues/206#issuecomment-403591369)

--> Solution: compile c code with CC (i.e., target `lib_target_test` etc.) and c++ code with CXX (i.e., targets `test` etc.)
--> Once this is resolved, remove `-Wno-error=deprecated` from `makefile`

dschlaep:
this is resolved for make target `make test test_run`, but not yet for `make test_severe test_run` or `make cov test_run`

Issue #206, opened 2018-07-04
Google test question


CaitlinA:
Has anything changed about the way we connect to or use the google test submodule?

Previously, I have cloned and/or checked out any branch on SOILWAT2 and google test would work after the 'make test' command.

I just clean cloned SOILWAT2 and 'make test' is not working. I used the command 'git submodule init' and 'git submodule update --remote'. I received this error:

> MacBook-Pro-4:SOILWAT2 candrews$ make test
c++  -Wall -Wextra -Werror -std=gnu++11		 -isystem googletest/googletest/include -Igoogletest/googletest \
			-pthread -c googletest/googletest/src/gtest-all.cc
In file included from googletest/googletest/src/gtest-all.cc:42:
googletest/googletest/src/gtest.cc:393:12: error: missing field 'owner_' initializer [-Werror,-Wmissing-field-initializers]
GTEST_API_ GTEST_DEFINE_STATIC_MUTEX_(g_linked_ptr_mutex);
           ^
googletest/googletest/include/gtest/internal/gtest-port.h:2096:80: note: expanded from macro 'GTEST_DEFINE_STATIC_MUTEX_'
     ::testing::internal::MutexBase mutex = { PTHREAD_MUTEX_INITIALIZER, false }
                                                                               ^
1 error generated.
make: *** [libgtest.a] Error 1

Maybe I am forgetting something. I can't find anything in the repository documentation.

dschlaep:
This is weird; I don't get this locally or see it on the CIs. We updated the submodule `googletest` a few days ago (commit https://github.com/DrylandEcology/SOILWAT2/commit/c159ebe9413b29aa6747b917cff63ec3166d4fdb): it was updated to commit https://github.com/google/googletest/commit/ba96d0b1161f540656efdaed035b3c062b60e006

Can you please check that you have the correct commit and a clean submodule by running
```
cd googletest/
git status
```
and making sure that you only get
```
HEAD detached at ba96d0b
nothing to commit, working tree clean
```

dschlaep:
@CaitlinA What is the status on this issue?

CaitlinA:
@dschlaep 

The googletest folder is pointing to the correct place

> MacBook-Pro-4:SOILWAT2 candrews$ cd googletest/
MacBook-Pro-4:googletest candrews$ git status
HEAD detached at ba96d0b
nothing to commit, working tree clean

I am still unable to run units tests and get the same error as above.

[...]

PR #209, opened 2018-07-11
Bugfix 206

dschlaep:
- address (but not resolve) issues #206 and #208
- adjustments so that the `clang` compiler works
    * set submodule `googletest` to specific commit 
    * specify specific warnings in the `makefile` to be ignored (instead of propagated to errors)


CaitlinA:
I think your edits look code and issue #206 can be closed.

I still receive warning for using clanging so #208 should remain open.

Issue #205, opened 2018-07-03
Linux machines have indirect memory leak errors


nip5:
Linux machines have indirect memory leak errors when the environment variable `ASAN_OPTIONS='detect_leaks=1'`. Thus the temporary solution is to instead run with that variable set to `0`. An instance of these errors:
```
Indirect leak of 232 byte(s) in 1 object(s) allocated from:
    #0 0x7f5d765d5602 in malloc (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x98602)
    #1 0x4a7a68 in Mem_Malloc(unsigned long, char const*) /home/natemccauslin/Desktop/DryLand Ecology/SOILWAT2/mymemory.c:104
    #2 0x4a7b0f in Mem_Calloc(unsigned long, unsigned long, char const*) /home/natemccauslin/Desktop/DryLand Ecology/SOILWAT2/mymemory.c:154
    #3 0x4ad922 in _newlayer() /home/natemccauslin/Desktop/DryLand Ecology/SOILWAT2/SW_Site.c:187
    #4 0x4b27ca in _read_layers /home/natemccauslin/Desktop/DryLand Ecology/SOILWAT2/SW_Site.c:553
    #5 0x4afd31 in SW_SIT_read() /home/natemccauslin/Desktop/DryLand Ecology/SOILWAT2/SW_Site.c:420
    #6 0x4a28ed in SW_CTL_read_inputs_from_disk() /home/natemccauslin/Desktop/DryLand Ecology/SOILWAT2/SW_Control.c:226
    #7 0x4a29ad in SW_CTL_obtain_inputs() /home/natemccauslin/Desktop/DryLand Ecology/SOILWAT2/SW_Control.c:255
    #8 0x408007 in Reset_SOILWAT2_after_UnitTest() test/sw_testhelpers.cc:45
    #9 0x40970c in TestBody test/test_SW_Carbon.cc:125
    #10 0x49246a in void testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) (/home/natemccauslin/Desktop/DryLand Ecology/SOILWAT2/sw_test+0x49246a)
[...]

PR #212, opened 2018-07-12
Better compilation and use in rSOILWAT2

### LINKED ISSUE #211: SOILWAT2 v4.1.2 has problems if compiled as rSOILWAT2 submodule
- function `RandNorm` returns a random variate with mean = 0 and sigma = 1 instead of `mean` and `stddev` respectively
- function `SW_OUT_read_onekey` doesn't use argument `period` if compiled with flag `RSOILWAT`

dschlaep:
- close #211

- function `SW_OUT_read_onekey`:  uses argument period if compiled with flag RSOILWAT to avoid `-Wunused-parameter`
- function `RandNorm`: if compiled with flag RSOILWAT, calls now function `rnorm` with two parameters `mean` and `sigma` from `Rmath.h` instead of `norm_rand`

### DIFF
--- a/.LSAN_suppr.txt
+++ b/.LSAN_suppr.txt
@@ -1,0 +1,7 @@
+# Suppression of memory leaks
+# https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizer#suppressions
+
+# These are known leaks that await fixing
+
+# https://github.com/DrylandEcology/SOILWAT2/issues/205
+leak:SW_VPD_construct
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,21 @@
 language: cpp
 
+# We want to run the `*_severe` debug and test targets; thus, we need
+# g++ >= 4.9 and clang++ >= 3.5
 compiler:
   - clang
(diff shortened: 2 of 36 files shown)


dschlaep:
This builds now fine as standalone, rSOILWAT2 (https://github.com/DrylandEcology/rSOILWAT2/pull/125) and STEPWAT2 (at least locally) and I suggest merging into master, but we have three documented and suppressed/ignored problems:
- memory leak (use of a suppression file): https://github.com/DrylandEcology/SOILWAT2/issues/205#issuecomment-404837304 --> @nip Can you please confirm that `ASAN_OPTIONS=detect_leaks=1 LSAN_OPTIONS=suppressions=.LSAN_suppr.txt make clean test_severe test_run` suppresses all current memory leaks for you as well?
- compile warning #208: treating c input as c++ --> @CaitlinA Can you please confirm that you no longer see this warning if you compile `make test`?
- compile warning #214: variable array length

nip5:
Yes, running that command suppresses the memory leaks and just leaves the variable array warnings.

Issue #226, opened 2018-09-06
Soilwat weather being affected by stepwat functions


chaukap:
Currently Stepwat is using random numbers to determine plant mortality and reproduction. The number of calls to rand() is variable depending on the input parameters of Stepwat. This means that random numbers that were being generated for Soilwat are instead being used by Stepwat. Therefore, for a given random seed the weather data from Soilwat will change depending on the input parameters to Stepwat. This is clearly not ideal behavior since weather cannot be kept constant for different grazing levels.

The problem comes from the fact that only one instance of the rand() function is created per program. Every call to rand by any function will affect every future call to rand. To solve the problem I would like to implement two new functions and add some functionality to rands.c and rands.h:
1: Declare the bucket array at the top of rands.c so that it can be filled with random numbers to be used by STEPWAT and any other coupled programs to avoid variable calls to rand().
2: Declare an array index field in rands.c. The index will be incremented every time STEPWAT requests a random number. This will cycle through the preselected numbers to give the impression that they are being created on the fly. 
3: Add function RandUni_deterministic(). This function will pick the next number from the bucket array and divide it by RAND_MAX to return a value between 0 and 1.
4: Add function RandUniRange_deterministic(int first, int last). This function will pick the next number from the bucket array and use it to generate a random number between first and last.
5: Add parameter to RandBeta(float aa, float bb, Bool deterministic). Rather than copy the RandBeta function and change two lines to create a deterministic version, I think it makes more sense to pass a Bool value that determines if RandBeta calls RandUni() or RandUni_deterministic().

The _deterministic versions of RandUni and RandUniRange should be used everywhere that a random number is requested outside of Soilwat. This will ensure that for a given seed Soilwat output will not be affected by anything outside of Soilwat.

I already have this code written and tested, and I plan on opening a new branch to submit it.

If anyone has any comments or questions about this issue please let me know. I am especially interested in getting feedback about adding a parameter to the RandBeta function.

kpalmqui:
@dschlaep @CaitlinA 

The random seed is not always working as expected in STEPWAT2. See description of this issue in https://github.com/DrylandEcology/STEPWAT2/issues/33#issuecomment-419180773. @chaukap has identified the problem and proposed a solution here (and also documented here https://github.com/DrylandEcology/STEPWAT2/issues/33#issuecomment-419180773), which needs to be implemented primarily in SOILWAT2.

[...]

Issue #210, opened 2018-07-11
Unit tests for `SWFlowTempTest.MainSoilTemperatureFunction_LyrMAX` fail intermittently


dschlaep:
Most of the time the unit tests pass well, but in rare cases the test for `SWFlowTempTest.MainSoilTemperatureFunction_LyrMAX` fails with

```
[ RUN      ] SWFlowTempTest.MainSoilTemperatureFunction_LyrMAX
test/test_SW_Flow_lib_temp.cc:569: Failure
Expected: (sTemp3[k]) < (100), actual: 999 vs 100
test/test_SW_Flow_lib_temp.cc:569: Failure
Expected: (sTemp3[k]) < (100), actual: 999 vs 100
test/test_SW_Flow_lib_temp.cc:569: Failure
Expected: (sTemp3[k]) < (100), actual: 999 vs 100
test/test_SW_Flow_lib_temp.cc:569: Failure
Expected: (sTemp3[k]) < (100), actual: 999 vs 100
test/test_SW_Flow_lib_temp.cc:569: Failure
Expected: (sTemp3[k]) < (100), actual: 999 vs 100
...
```

This is easy to ignore locally but simply re-running the unit tests; however, this is not suitable for our CI checks.

--> My guess is that this is because of the use of random number generator in the initialization of the test variables which produces values that break the function (unexpectedly) in these rare cases:
```
    for (i = 0; i < nlyrs2; i++) {
      bDensity2[i] = fmaxf(RandNorm(1.,0.5), 0.1); // greater than 0.1
      fc2[i] = fmaxf(RandNorm(1.5, 0.5), 0.1); // greater than 0.1
...
```
which was introduced by commit https://github.com/DrylandEcology/SOILWAT2/commit/ae0b6ec02e9e4fc9eae5abeb9b9f71aa40e450c6

PR #233, opened 2018-09-18
Incorporate pcg rng

dschlaep:
Switch to a modern random number generator which allows multiple streams. We chose `PCG` (http://www.pcg-random.org/) which 

> is a family of simple fast space-efficient statistically good algorithms for random number generation

### DIFF
--- a/SW_Model.c
+++ b/SW_Model.c
@@ -48,12 +48,15 @@
 #include "SW_SoilWater.h"  /* for setup_new_year() */
 #include "SW_Times.h"
 #include "SW_Model.h"
+#include "pcg/pcg_basic.h"
 
 /* =================================================== */
 /*                  Global Variables                   */
 /* --------------------------------------------------- */
 extern SW_SITE SW_Site; /* for reset attribute */
 SW_MODEL SW_Model; /* declared here, externed elsewhere */
+
+pcg32_random_t markov_rng;
 
 /* =================================================== */
 /*                Module-Level Variables               */
@@ -90,7 +93,7 @@ void SW_MDL_construct(void) {
 
 #ifndef STEPWAT
 	/* already set by user-provided seed in steppe */
-	RandSeed(0);
+	RandSeed(0,&markov_rng);
 #endif
 }
 
--- a/SW_Model.c
+++ b/SW_Model.c
@@ -48,7 +48,7 @@
 #include "SW_SoilWater.h"  /* for setup_new_year() */
 #include "SW_Times.h"
 #include "SW_Model.h"
-#include "pcg/pcg_basic.h"
+
 
 /* =================================================== */
 /*                  Global Variables                   */
--- a/SW_Model.c
+++ b/SW_Model.c
@@ -95,7 +95,7 @@ void SW_MDL_construct(void) {
 
 #ifndef STEPWAT
 	/* already set by user-provided seed in steppe */
-	RandSeed(Globals.randseed,&markov_rng);
+	RandSeed(0,&markov_rng);
 #endif
 }
 
--- a/SW_Model.c
+++ b/SW_Model.c
@@ -56,7 +56,9 @@
 extern SW_SITE SW_Site; /* for reset attribute */
 SW_MODEL SW_Model; /* declared here, externed elsewhere */
 
-pcg32_random_t markov_rng;
+#ifndef STEPWAT
+  pcg32_random_t markov_rng;
+#endif
 
 /* =================================================== */
 /*                Module-Level Variables               */
@@ -93,7 +95,7 @@ void SW_MDL_construct(void) {
 
 #ifndef STEPWAT
 	/* already set by user-provided seed in steppe */
-	RandSeed(0,&markov_rng);
+	RandSeed(Globals.randseed,&markov_rng);
 #endif
 }
 
--- a/SW_Model.c
+++ b/SW_Model.c
@@ -56,10 +56,6 @@
 extern SW_SITE SW_Site; /* for reset attribute */
 SW_MODEL SW_Model; /* declared here, externed elsewhere */
 
-#ifndef STEPWAT
-  pcg32_random_t markov_rng;
-#endif
-
 /* =================================================== */
 /*                Module-Level Variables               */
 /* --------------------------------------------------- */
@@ -92,11 +88,6 @@ void SW_MDL_construct(void) {
 		SW_Model.newperiod[pd] = swFALSE;
 	}
 	SW_Model.newperiod[eSW_Day] = swTRUE; // every day is a new day
-
-#ifndef STEPWAT
-	/* already set by user-provided seed in steppe */
-	RandSeed(0,&markov_rng);
-#endif
 }
 
 void SW_MDL_deconstruct(void)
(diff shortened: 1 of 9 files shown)


rands.c
```diff
@@ -19,38 +20,30 @@ long _randseed = 0L;
...
- be removed in the near future.
+  \param Seed is the initial state of the system. 0 indicates a random seed.
+  \param pcg_rng is the random number generator to set.
 
- \param seed The seed.
-
- cwb - 6/27/00
 */
-void RandSeed(signed long seed) {
+void RandSeed(signed long seed, pcg32_random_t* pcg_rng) {
```
dschlaep: The body of this function must have a `#ifdef RSOILWAT` case -- c code compiled for the R package `rSOILWAT2` must use the R API RNGs (see https://cran.r-project.org/doc/manuals/R-exts.html#Writing-portable-packages and https://cran.r-project.org/doc/manuals/R-exts.html#Random-numbers).

rands.c
```diff
@@ -60,137 +53,20 @@ void RandSeed(signed long seed) {
...
-
-	\return double. A value between 0 and 1.
+  \parameters: pcg32_random_t* that specifies which stream to use.
+ 
+  \returns: a double between 0 and 1.
 */
-double RandUni_fast(void) {
-	static double y;
-
-	#ifdef RSOILWAT
```
dschlaep: You cannot deleted this `#ifdef RSOILWAT` case -- c code compiled for the R package `rSOILWAT2` must use the R API RNGs (see https://cran.r-project.org/doc/manuals/R-exts.html#Writing-portable-packages and https://cran.r-project.org/doc/manuals/R-exts.html#Random-numbers).

rands.c
```diff
@@ -60,137 +53,20 @@ void RandSeed(signed long seed) {
...
-		i = max(BUCKETSIZE -1, min( i,0));
-		/*  i = (i > BUCKETSIZE -1 ) ? BUCKETSIZE -1 : i; */
-
-		/* snatch a random number and replace it */
-		y = bucket[i];
-		bucket[i] = ((double) ix1 + (double) ix2 * rm2) * rm1;
-	#endif
-
-	return y;
+double RandUni(pcg32_random_t* pcg_rng) {
```
dschlaep: The code needs the now deleted:
```
#ifdef RSOILWAT
		GetRNGstate();
		y = unif_rand();
		PutRNGstate();
```

rands.c
```diff
@@ -234,12 +110,11 @@ int RandUniRange(const long first, const long last) {
 	}
 
 	r = l - f + 1;
-	return (long) (RandUni() * r) + f;
+	return (long) (RandUni(pcg_rng) * r) + f;
```
dschlaep: Please call `pcg32_boundedrand_r`; see http://www.pcg-random.org/using-pcg-c.html#pcg32-boundedrand-r-rngptr-bound for an explanation.

SW_Model.c
```diff
@@ -48,13 +48,18 @@
...
 #include "SW_Model.h"
+#include "pcg/pcg_basic.h"
 
 /* =================================================== */
 /*                  Global Variables                   */
 /* --------------------------------------------------- */
 extern SW_SITE SW_Site; /* for reset attribute */
 SW_MODEL SW_Model; /* declared here, externed elsewhere */
 
+#ifndef STEPWAT
```
dschlaep: Why wrap this in `#ifndef STEPWAT`? -- this is confusing; extern `markov_rng` on the STEPWAT2 side if you need it there instead of two different places to define this variable.

rands.c
```diff
@@ -19,38 +20,30 @@ long _randseed = 0L;
 
 /*****************************************************/
 /**
-  \fn void RandSeed(signed long seed)
-  \brief Resets the random number seed.
+  \brief Sets the random number seed. This should only be done ONCE per iteration.
```
dschlaep: `iteration` is `STEPWAT2` jargon which a `SOILWAT2` user does not understand. Provide context.

rands.c
```diff
@@ -19,38 +20,30 @@ long _randseed = 0L;
...
+  else {
+    //Seed with a specific value.
+    pcg32_srandom_r(pcg_rng, (int) seed, stream);
+  }
+
+  //Increment the stream so no two generators have the same sequence.
+  stream++;
 
 #ifndef RSOILWAT
 #if RAND_FAST
```
dschlaep: Please remove `RAND_FAST` -- we eliminated `RandUni_fast` and `RandUni_good` and thus no longer need to select between them.

SW_Model.c
```diff
@@ -90,7 +95,7 @@ void SW_MDL_construct(void) {
 
 #ifndef STEPWAT
 	/* already set by user-provided seed in steppe */
-	RandSeed(0);
+	RandSeed(0,&markov_rng);
```
dschlaep: As suggested in my https://github.com/DrylandEcology/SOILWAT2/issues/226#issuecomment-419931492, the markov weather generator would ideally obtain its own seed. That means a new input for `SOILWAT2`.

dschlaep: I don't understand why you put all of this code related to `markov_rng` into the module `SW_Model.c`; it is not used by this code. Why not put it instead in `SW_Markov.c`, define `markov_rng` in that module, and set the seed by the function `SW_MKV_construct()`?

rands.c
dschlaep: Would be great to add unit test cases to `test/test_rands.cc` to demonstrate how to use multiple streams, how the set them up with `RandSeed`, and what to expect!

rands.c
chaukap: @dschlaep If I call pcg32_boundedrand_r() will I need to add an 
#ifdef RSOILWAT case to this function like RandUni()?

dschlaep:
Yes, there needs to be such a case for rSOILWAT2's code (see https://cran.r-project.org/doc/manuals/R-exts.html#Distribution-functions):
 
```
...

#ifdef RSOILWAT
	GetRNGstate();
	y = runif(first, last);
	PutRNGstate();

#else
	...

#endif
```

rands.c
```diff
@@ -19,183 +20,124 @@ long _randseed = 0L;
...
+  PutRNGstate();
+
+#else
+  // get a random unsigned int and cast it to a signed int.
+  number = (int) pcg32_random_r(pcg_rng);
+  // negative values are possible. This takes the absolute value.
+  number = fabs(number); 
+  //printf("%f\n", number/(double)RAND_MAX);
+  // divide by RAND_MAX to get a value between 0 and 1.
+  number = number / (double)RAND_MAX;
```
dschlaep: `pcg32_random_r / RAND_MAX` is not guaranteed to be as expected. `RAND_MAX` is defined in `stdlib.h` and its value is implementation specific, albeit at least 32767.

`pcg32_random_r`, however, produces an integer between 0 and 2^32 (see, http://www.pcg-random.org/using-pcg-c-basic.html).
To correctly get a value between 0 and 1, you need to follow the `pcg` instructions, i.e., 
```
double d = ldexp(pcg32_random_r(&myrng), -32);
```

dschlaep:
again #210

Issue #237, opened 2018-10-02
Add pcg rngs to google test files


chaukap:
Out contiguous integration tests are failing because the google tests are not updated with the new rand function calls. RNGs need to be added to each google test.

chaukap:
This issue has been closed with pull request #238

PR #238, opened 2018-10-02
Added pcg rngs to gtests. (issue #237)

chaukap:
Each test that calls a random function now has its own pcg32_random_t RNG. The RNG is then seeded with a 0 so that the output will always be different.

### DIFF
--- a/test/test_SW_Flow_Lib.cc
+++ b/test/test_SW_Flow_Lib.cc
@@ -34,6 +34,7 @@
 #include "../SW_Weather.h"
 #include "../SW_Markov.h"
 #include "../SW_Sky.h"
+#include "../pcg/pcg_basic.h"
 
 #include "../SW_Flow_lib.h"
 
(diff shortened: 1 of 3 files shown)


chaukap:
I have added RNGs to all tests that need them, but I am still failing the Travis CI build. The test that I am failing is:

[  FAILED  ] SWFlowTempTest.MainSoilTemperatureFunction_LyrMAX (7 ms)

Using the clang compiler. Im not quite sure why this test failed, because I only added the rng without changing anything else.

chaukap:
@dschlaep would you mind taking a look at how I integrated RNGs into the google tests? Everything looked good but there is still one test failing in Travis with clang.

dschlaep:
@chaukap I'm sorry; you are correct, the failure of this test had nothing to do with what you did. I documented it in issue #210 but no-one has fixed it yet. Usually, a re-build on the CI will succeed -- which I did.

I will review by tomorrow or Thursday -- is that ok?

chaukap:
@dschlaep Sounds great! I'll pull this into the incorporate_pcg_rng branch so that you can review the whole thing.

PR #240, opened 2018-11-28
Fix `SW_NFILES`

dschlaep:
correct number of files is 22 and not 23


dschlaep:
this is addressing https://github.com/DrylandEcology/rSOILWAT2/issues/130

PR #241, opened 2018-12-10
Feature interception

dschlaep:
New interception equations


dschlaep:
this PR will become v5.0.0 (because user input files `veg.in` and `climate.in` change)

Issue #231, opened 2018-09-14
Unit test for `SWSoilWaterTest` can fail with division by zero


dschlaep:
e.g., seen with https://travis-ci.org/DrylandEcology/SOILWAT2/jobs/428772098 for commit 00ab61da8bd4bea0628b0f4a9de913661d5beb3d

>[ RUN      ]SWSoilWaterTest.SWSWCbulk2SWPmatric
SW_SoilWater.c:987:25: runtime error:division by zero
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior SW_SoilWater.c:987:25 in

PR #242, opened 2019-02-13
Divide by zero bug fix, issue #231

nip5:
Addressing issue#231.

* Note that this will decrease code coverage slightly as this issue is related to a `CONSTANT` value not being visible.

### DIFF
--- a/SW_SoilWater.c
+++ b/SW_SoilWater.c
@@ -984,12 +984,13 @@ RealD SW_SWCbulk2SWPmatric(RealD fractionGravel, RealD swcBulk, LyrIndex n) {
 			theta1 = 0.0;
 		else
 			theta1 = (swcBulk / lyr->width) * 100. / (1. - fractionGravel);
-		swp = lyr->psisMatric / powe(theta1/lyr->thetasMatric, lyr->bMatric) / BARCONV;
-	} else {
+    if (GT(BARCONV, 0.0))
+      swp = lyr->psisMatric / powe(theta1/lyr->thetasMatric, lyr->bMatric) / BARCONV;
+    else
+      LogError(logfp, LOGFATAL, "Attempt to divide by 0, SWP value may not be accurate.");
+	} else
 		LogError(logfp, LOGFATAL, "Invalid SWC value (%.4f) in SW_SWC_swc2potential.\n"
 				"    Year = %d, DOY=%d, Layer = %d\n", swcBulk, SW_Model.year, SW_Model.doy, n);
-	}
-
 	return swp;
 }
 


dschlaep:
@nip5 @CaitlinA What was this for? Who approved this PR? Where are the new unit tests that demonstrate that this works?

Has anyone bothered to figure out what `BARCONV` is? The person who would look would find that it is `#define BARCONV     1024.` in `SW_Defines.h`. So, of course it is always larger than 0 and of course the check added with this commit is utterly useless and does not fix #231 at all.

Please undo.

btw, according to any dictionary including ODE, NOAD, and Merriam-Webster, the word `visable` does not exists and it is not obvious to me what it should mean

nip5:
Okay I will undo this, I made the mistake of thinking that since it passes all the CIs that I could auto merge it but have since talked to Caitlin about it, now I know. `BARCONV `doesn't always seem to be visible, if it were than #231 would not be an issue because to your point, it should always be 1024.

`Visable` was obviously a typo and should have been caught by me; it is clear what is meant here. I will revert the the push.

btw, it is never "useless" to add checks to your code that verify that a value exists before using it.

dschlaep:
`BARCONV` is a globally defined constant; it is replaced by the pre-compiler to its constant value, here 1024. Thus your check is indeed useless because 1024 is always larger than 0. `BARCONV` cannot change its value at runtime.

I still don't get what you mean with whether or not `BARCONV` is "visible" -- of course it is "visible" because it is defined globally...
> visible | ˈvizəb(ə)l |
adjective
1 able to be seen: the church spire is visible from miles away.
• Physics (of light) within the range of wavelengths to which the eye is sensitive.
• able to be perceived or noticed easily: a visible improvement.
• in a position of public prominence: a highly visible member of the royal entourage.
2 relating to imports or exports of tangible commodities: the visible trade gap.

Maybe you attempted to write "divisible"? Even though the meaning of that could be guessed, it is still misleading because we are not operating with whole numbers but with floating point numbers (e.g., https://en.wikipedia.org/wiki/Divisor). As the title of #231 states, the problem is not "divisibility", but instead "division by zero", i.e., #231 exists because a denominator is zero in its floating point representation (e.g., https://en.wikipedia.org/wiki/Division_by_zero). Here, the denominator(s) of the line
```
swp = lyr->psisMatric / powe(theta1/lyr->thetasMatric, lyr->bMatric) / BARCONV;
```
in question is/are obviously `powe(theta1/lyr->thetasMatric, lyr->bMatric)` and/or `lyr->thetasMatric` because it/they can become zero at runtime depending on the values of `theta1`, `sand`, etc.

nip5:
I know all that, what I am saying is that in a situation where the dependency on `SW_Defines.h` fails to load in `SW_SoilWater.c`, `BARCONV` would not be registered correctly by `SW_SoilWater.c` (thus is not "visible") and thus would cause issues. This justifies the need for a check and if indeed the problem is not related to `BARCONV `as you suggest, than my check would not have fixed the error, yet it did.

dschlaep:
`SOILWAT2` doesn't compile if `SW_Defines.h` doesn't "load"; thus your check will never be executed because there is no executable to run the code.

You have not demonstrated that your "fix" fixed #231.

PR #264, opened 2019-07-11
Enhancement 262 biomassunits

dschlaep:
- corrected output when run with `STEPWAT2`

- new output target `BIOMASS` provides output for
  * cover (%) for trees, shrubs, forbs, grasses
  * biomass (g/m2 as component of total) for total, trees, shrubs, forbs, grasses, and litter
  * live biomass (g/m2 as component of total) total, trees, shrubs, forbs, grasses

--> user visible change to the input file `outsetup.in` which gained a new row!

### DIFF
--- a/SW_Output_get_functions.c
+++ b/SW_Output_get_functions.c
@@ -246,7 +246,7 @@ void get_co2effects_agg(OutPeriod pd) {
 
 	if (print_IterationSummary) {
 		sw_outstr_agg[0] = '\0';
-		format_IterationSummary(p, psd, pd, 2 * NVEGTYPES);
+		format_IterationSummary(p, psd, pd, ncol_OUT[eSW_CO2Effects]);
 	}
 }
 #endif
@@ -395,7 +395,7 @@ void get_biomass_agg(OutPeriod pd) {
 
 	if (print_IterationSummary) {
 		sw_outstr_agg[0] = '\0';
-		format_IterationSummary(p, psd, pd, 16);
+		format_IterationSummary(p, psd, pd, ncol_OUT[eSW_Biomass]);
 	}
 }
 #endif
@@ -465,7 +465,7 @@ void get_estab_agg(OutPeriod pd)
 
 	if (print_IterationSummary) {
 		sw_outstr_agg[0] = '\0';
-		format_IterationSummary(p, psd, pd, v->count);
+		format_IterationSummary(p, psd, pd, ncol_OUT[eSW_Estab]);
 	}
 }
 #endif
@@ -516,7 +516,7 @@ void get_temp_agg(OutPeriod pd)
 
 	if (print_IterationSummary) {
 		sw_outstr_agg[0] = '\0';
-		format_IterationSummary(p, psd, pd, 4);
+		format_IterationSummary(p, psd, pd, ncol_OUT[eSW_Temp]);
 	}
 }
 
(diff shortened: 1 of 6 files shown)


SW_Output_get_functions.c
```diff
@@ -188,135 +188,220 @@ void get_none(OutPeriod pd)
...
+		sw_outstr_agg[0] = '\0';
+		format_IterationSummary(p, psd, pd, ncol_OUT[eSW_CO2Effects]);
+	}
+}
+#endif
+
+
+
+
+//------ eSW_Biomass
```
kpalmqui: delete extra spaces?

SW_Output_get_functions.c
```diff
@@ -188,135 +188,220 @@ void get_none(OutPeriod pd)
...
+	do_running_agg(p, psd, iOUT(i + NVEGTYPES, pd), Globals.currIter,
+		biomass_total);
+	i += NVEGTYPES + 1;
+	ForEachVegType(k) {
+		do_running_agg(p, psd, iOUT(i + k, pd), Globals.currIter,
+			vo->veg[k].biomass * v->veg[k].cov.fCover);
+	}
+	do_running_agg(p, psd, iOUT(i + NVEGTYPES, pd), Globals.currIter,
+		litter_total);
+
```
kpalmqui: why did you choose not to export litter for each veg type here?

SW_Output_get_functions.c
dschlaep: addressed by 987b972082b2f5bde56c3dc486a8bc911b7f6073

SW_Output_get_functions.c
dschlaep: @kpalmqui I thought that adding four additional columns of output does not add new information and only results in the output files being even larger. 

This is because litter does not change with simulation time in `SOILWAT2` standalone runs, i.e., inputs = outputs --> no use for such runs. 

In `STEPWAT2` runs, litter changes from year-to-year, but `STEPWAT2` currently does not vary amounts of litter between functional types, i.e., total litter output is sufficient for `STEPWAT2` runs. See `STEPWAT2` litter inputs (https://github.com/DrylandEcology/STEPWAT2/blob/master/testing.sagebrush.master/Stepwat_Inputs/Input/sxw/sxwprod_v2.in) and see corresponding variable `_prod_litter` which does not vary by vegetation type or by resource group). However, I would add litter output for each vegetation type if you still want to?

dschlaep:
> @kpalmqui One issue is now the SOILWAT2 output is being written to sw_output within a new folder /Stepwat_Inputs/Output_ref

--> The output directory for the `SOILWAT2` output file that are produced as part of a `STEPWAT2`run is determined by the relative paths provided as inputs in the file `https://github.com/DrylandEcology/STEPWAT2/blob/master/testing.sagebrush.master/Stepwat_Inputs/Input/sxw/files_SOILWAT2.in`. I double checked and they have not been changed on any of the currently active `STEPWAT2` branches.
--> My guess is that a local copy of this file exists that has modified this path? Or, that `Output_ref` is a leftover from running the script I suggested in https://github.com/DrylandEcology/STEPWAT2/pull/337/commits/af5c481f1930c396f61230002d6e41708a004e9d?

SW_Output_get_functions.c
kpalmqui: @dschalep no your decision makes sense given the above explanation. thanks!

kpalmqui:
@dschalep - not sure, but I tested enhancement_262_biomassunits within testing_bugfixes_336_346 and this issue disappeared.

PR #289, opened 2021-11-17
Updated approach for global variables

dschlaep:
Updated approach for global variables

- previously, defined/implemented global variables in one source file & used extern declaration by each other source files that referenced a global variable
--> problem: repeated declarations

- new approach (https://stackoverflow.com/questions/1433204/how-do-i-use-extern-to-share-variables-between-source-files?rq=1): 
* define/implement global variables in one source file
* extern declare these variables in the header file belonging to that source file
* include the header file by each other source files that referenced a global variable

- see also https://github.com/DrylandEcology/STEPWAT2/issues/512 (Compilation fails with gcc >= v10 due to duplicate symbols)


kpalmqui (approved):
@dschlaep looks good, sorry for the delay.

PR #293, opened 2022-01-12
Improved percolation and minimum SWC

### LINKED ISSUE #292: Soil moisture can drop too fast because percolation rate is too fast
Simulated soil moisture can drop to very low values, particularly if expressed as soil water potential (top right panel). This appears to be because unsaturated percolation doesn't sufficiently decrease with decreasing soil moisture (left panels).

Percolation_vs_Moisture_daily

Red lines indicate soil moisture at -100, -1.5, and -0.033 MPa. Here, minimum moisture was set to -100 MPa.

### LINKED ISSUE #295: Soil moisture can drop below minimum levels
Observed in
```
WaterBalance.WithHighGravelVolume
WB (1980-124) sl=0: swc_min(0.003699) vs. swc(0.003008)
```

dschlaep:
* Lower limit for `swc_min` updated to -30 MPa.
* Unsaturated percolation rate is now adjusted for `swc_min` (close #292).
* New water balance checks make sure that `swc_min` <= `swc` <= `swc_sat` (close #295).
  * Bare-soil evaporation no longer evaporates soil moisture held below `swc_min`.
  * Transpiration no longer evaporates soil moisture held below `swc_min`.
  * Hydraulic redistribution no longer moves soil moisture held below `swc_min`.


kpalmqui (approved):

dschlaep:
* Lower limit for `swc_min` updated to -30 MPa
  (previously -3 MPa which was too high for dryland ecosystems).


* Unsaturated percolation rate is now adjusted for `swc_min`.
  Previously, soils were drained too fast at low moisture levels because
  the rate, incorrectly, didn't reach 0 when soil moisture approached `swc_min`
  (#292 "Soil moisture can drop too fast because percolation rate is too fast").


* New water balance checks make sure that `swc_min` <= `swc` <= `swc_sat`
  (#295 "Soil moisture can drop below minimum levels").

* Bare-soil evaporation no longer evaporates soil moisture held below `swc_min`.
  Previously, evaporation could reduce `swc` below `swc_min` because
  the bare-soil evaporation limit "half wilting point" was not checked to be
  larger than `swc_min`. Now, the limit is forced to be larger than `swc_min`
  and larger than the hygroscopic point at -10 MPa
  (#295 "Soil moisture can drop below minimum levels").

Some more detailed notes:

* Transpiration no longer evaporates soil moisture held below `swc_min`.
  Previously, transpiration could reduce `swc` below `swc_min` because
  the transpiration limits "critical SWPs" were not checked to be
  larger than `swc_min`. Now, the limits are forced to be larger than `swc_min`
  (#295 "Soil moisture can drop below minimum levels").


* Hydraulic redistribution no longer moves soil moisture held below `swc_min`.
  (#295 "Soil moisture can drop below minimum levels").
  1) Previously, a narrow soil layer could "overcommit" moisture
     (which that layer does not contain, e.g., > 80x in some observed cases)
     because previous code did not adjust original equations which were
     developed for soil layers of equal width. Now, recipient relative rooting
     contribution is adjusted to what it would be if it were the same width as
     a narrower source layer.
  2) Now, each source/recipient redistribution is limited to source moisture
     above a set minimum level.
  3) Previously, code failed in the attempt to limit net redistribution for a
     layer to moisture held above a set minimum level.
     Now, net redistribution is correctly limited with
     (i) correct scaling equations
     (ii) reductions in source moisture (instead of all)
     (iii) a loop over layers and repeated checks if net redistribution requires
           adjustments
           (previously, only one round of adjustment was made and did
           not consider that restricting redistribution in one layer would
           affects net redistribution in all recipient layers).
  4) The set minimum moisture level above which hydraulic redistribution
     may move water is now set to vegetation type specific critical SWP or
     wilting point (whichever is lower), here between -2 and -3.9 MPa.
     Previously, dry sites with low water holding capacity and if minimum moisture levels were set to a low value (e.g., below -30 MPa), then
     redistribution of near identical amounts of water back-and-forth among some layers on consecutive days was rarely observed and led to a "zig-zag" pattern in soil moisture. However, users should be on the look-out for a repeat of this artifact since the code does not actively monitor for this situation.

PR #323, opened 2022-08-02
Specify type of soil density inputs

### LINKED ISSUE #280: Soil density: input either for < 2 mm fraction or for whole soil?
Consider adding a flag to indicate whether inputs reflect soil density of the < 2 mm fraction or soil density of the whole soil (with rocks/gravel).

And then calculate internally what SOILWAT2 requires, e.g., with the already existing function `calculate_soilBulkDensity()`

dschlaep:
- close #280 (Soil density: input either for < 2 mm fraction or for whole soil)

* new behavior: user specifies whether input soil density represents matric or bulk density; code then determines both matric and bulk density values from inputs
* previous behavior: input soil density was treated as matric density and converted to bulk density

### DIFF
--- a/SW_Site.c
+++ b/SW_Site.c
@@ -110,7 +110,7 @@ static void _read_layers(void) {
 	FILE *f;
 	LyrIndex lyrno;
 	int x, k;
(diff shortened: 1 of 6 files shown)


dschlaep:
this should be a PR into `release/devel_v7.0.0`!

PR #338, opened 2023-01-09
Daily variables for atmospheric demand

### LINKED ISSUE #341: Daily variables for atmospheric demand
Handle daily inputs of SOILWAT2 daily driving variables

- required daily inputs
    - tmin/tmax/tmean
    - ppt
- mean monthly or daily inputs
    - Wind speed (sfcWind)​​: (i) wind speed, or (ii) east/north wind components
    - Relative humidity (hurs): (i) min/max relative humidity, or (ii) relative humidity, or (iii) specific humidity, or (iv) vapor pressure, or (v) dewpoint temperature
    - Cloud cover (clt), optional if rsds daily inputs available
- optional daily inputs
    - Actual vapor pressure (e_a): (i) vapor pressure, or (ii) dewpoint temperature, or (iii) min/max relative humidity, or (iv) relative humidity, or (v) specific humidity
    - Downward surface shortwave radiation (rsds) = H_gh: flux density as mean over 24 hours or as mean over daylight period

dschlaep:
Handle daily inputs of SOILWAT2 daily driving variables

- tmin/tmax/tmean
- ppt
- Wind speed (sfcWind)​​
- Relative humidity (hurs)
- Actual vapor pressure (e_a)
- Cloud cover (clt)
- Downward surface shortwave radiation (rsds) = H_gh

-> close #341


src/SW_Weather.c
```diff
@@ -1065,17 +1350,99 @@ void SW_WTH_setup(void) {
...
 
 		lineno++;
 	}
 
+    // Check if monthly flags have been chosen to override daily flags
+    w->has_sfcWind = (w->use_windSpeedMonthly) ? swFALSE : w->has_sfcWind;
+    w->has_windComp = (w->use_windSpeedMonthly) ? swFALSE : w->has_windComp;
+
+    w->has_hurs = (w->use_relHumidityMonthly) ? swFALSE : w->has_hurs;
+    w->has_hurs2 = (w->use_relHumidityMonthly) ? swFALSE : w->has_hurs2;
```
dschlaep, line 1365: If a run uses monthly humidity inputs, then all humidity related daily inputs should be turned off (otherwise vp and rH may end up being inconsistent), i.e.,
```
    if (w->use_relHumidityMonthly) {
      w->has_hurs = swFALSE;
      w->has_hurs2 = swFALSE;
      w->has_huss = swFALSE;
      w->has_vp = swFALSE;
    }
```

include/SW_Weather.h
```diff
@@ -153,11 +157,23 @@ typedef struct {
...
     scale_wind[MAX_MONTHS],
-    scale_rH[MAX_MONTHS];
+    scale_rH[MAX_MONTHS],
+    scale_actVapPress[MAX_MONTHS],
+    scale_shortWaveRad[MAX_MONTHS];
 	char name_prefix[MAX_FILENAMESIZE - 5]; // subtract 4-digit 'year' file type extension
 	RealD snowRunoff, surfaceRunoff, surfaceRunon, soil_inf, surfaceAvg;
 	RealD snow, snowmelt, snowloss, surfaceMax, surfaceMin;
 
+  Bool use_cloudCoverMonthly, use_windSpeedMonthly, use_relHumidityMonthly,
```
dschlaep, line 167: I suggest that we rename `use_relHumidityMonthly` to `use_humidityMonthly` -- because we need to be consistent, i.e., all humidity aspects are derived (consistently) from monthly inputs or none; see also comment to line 1364 `w->has_hurs = (w->use_relHumidityMonthly) ? swFALSE : w->has_hurs;` etc.

src/SW_Weather.c
```diff
@@ -527,6 +548,16 @@ void readAllWeather(
               weather_prefix
             );
         }
+
+        // Check to see if actual vapor pressure needs to be calculated
+        if(!SW_Weather.has_vp && SW_Weather.use_relHumidityMonthly) {
```
dschlaep, line 553: - remove `SW_Weather.has_vp` from the condition `if(!SW_Weather.has_vp && SW_Weather.use_relHumidityMonthly) {` -- either we use monthly humidity inputs and then we need to calculate vapor pressure -- or, we use daily inputs for humidity related variables and then vapor pressure is calculated inside `_read_weather_hist()` in any case
- also we cannot calculate vapor pressure here if needed because of monthly humidity inputs because of the situation when the weather generator is in use and `allHist[yearIndex]->temp_avg[day]` is missing;
the best location that I can currently think of is inside `finalizeAllWeather()` just after the call to `generateMissingWeather()`

src/SW_Weather.c
```diff
@@ -527,6 +548,16 @@ void readAllWeather(
...
+
+            for(day = 0; day < MAX_DAYS; day++) {
+                allHist[yearIndex]->actualVaporPressure[day] =
+                                actualVaporPressure1(allHist[yearIndex]->r_humidity_daily[day],
+                                                     allHist[yearIndex]->temp_avg[day]);
+            }
+        }
     }
 }
```
dschlaep, line 563: We will need to expand functionality of `generateMissingWeather()` so that `method = 1` (LOCF = last value carried forward) includes all variables in `allHist`, i.e., add equivalent code to
```
          allHist[yearIndex]->temp_max[day] = missing_Tmax ?
            yesterdayMax :
            allHist[yearIndex]->temp_max[day];
```
for the new variables in `allHist`, etc

dschlaep (commented):
I added some reviews; most are minor and address clarity and documentation. Others suggest that the logic of priorities of daily inputs should be double checked against our specifications; that `generateMissingWeather()` should gain LOCF functionality for all of `allHist`; and any "allWeather" functionality needs to be free from global variables (requirement for STEPWAT2)

src/SW_Weather.c
N1ckP3rsl3y: Would we not need to add `SW_Weather.has_hurs` in place of `!SW_Weather.has_vp`? Since we calculate actual vapor pressure with `actualVaporPressure1()` then as well. But this will need to be backed up in some way making sure something of higher priority (e.g., min/max hurs, ...) is not given.

dschlaep: `hurs` is a daily input and all cases of calculating `vp` from daily input needs to happen by `_read_weather_hist()` [see also https://github.com/DrylandEcology/SOILWAT2/pull/338#discussion_r1066200121]-- calculating `vp` is needed here (respectively later [see second bullet point of my comment]) if we use monthly humidity values

tests/gtests/test_SW_Flow_Lib_PET.cc
```diff
@@ -977,14 +986,16 @@ namespace
       // Inputs
       RHs[] = {0, 34, 56, 79, 100},
       // Expected PET
-      expected_pet_RHs[] = {0.2267, 0.2123, 0.1662, 0.1128, 0.0612};
+      expected_pet_RHs[] = {0.2267, 0.2123, 0.1662, 0.1128, 0.0612},
+      // Input actual vapor pressure
+      e_a_in[] = {0.0, 1.077044, 1.773956, 2.502544, 3.167778};
```
dschlaep, line 991: please also calculate here vapor pressure from temperature and humidity

tests/gtests/test_SW_Weather.cc
```diff
@@ -686,5 +686,120 @@ namespace {
...
+        int yearIndex = 0, midJanDay = 14;
+
+        /* Test if monthly values are not being used */
+
+        SW_WTH_setup();
+
+        SW_Weather.use_humidityMonthly = swTRUE;
+        SW_Weather.use_windSpeedMonthly = swTRUE;
+        SW_Weather.use_cloudCoverMonthly = swTRUE;
+        SW_Weather.has_hurs = swTRUE;
```
dschlaep, line 709: Why `SW_Weather.has_hurs = swTRUE;` if we only use monthly values here?

src/SW_Weather.c
```diff
@@ -511,18 +512,38 @@ void readAllWeather(
...
+        year = yearIndex + startYear;
 
         // Set all daily weather values to missing
         _clear_hist_weather(allHist[yearIndex]);
 
+        // Update yearly day/month information needed when interpolating
+        // cloud cover, wind speed, and relative humidity if necessary
+        Time_new_year(year);
+
+        if(SW_Weather.use_cloudCoverMonthly) {
```
dschlaep, line 530: `readAllWeather()` still uses globals "SW_Weather" and "SW_Sky" -- see earlier comment https://github.com/DrylandEcology/SOILWAT2/pull/338#discussion_r1066025836

src/SW_Weather.c
```diff
@@ -1065,17 +1350,99 @@ void SW_WTH_setup(void) {
...
 		LogError(logfp, LOGFATAL, "%s : Too few input lines.", MyFileName);
 	}
+
+    // Calculate value indices for `allHist`
+
+    // Default n_input_forcings to 0
+    w->n_input_forcings = 0;
+
+        // Loop through MAX_INPUT_COLUMNS
+    for(columnNum = 3; columnNum < MAX_INPUT_COLUMNS; columnNum++)
```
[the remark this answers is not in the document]
dschlaep: Monthly flags? Why mix monthly and daily flags in `inputFlags[]`?
```
        // The loop starts at 3 due to "inputFlags[]" containing monthly flags
        // in the first three slots (indices 0, 1, and 2)
```

I assumed `inputFlags[]` would be flags for daily inputs and indices 0 and 1 would represent daily max and min temperature inputs (i.e., `has_temp2`) and index 3 would represent daily precipitation inputs (i.e., `has_ppt`) -- as listed in "weathsetup.in"

src/SW_Weather.c
dschlaep: Nice job with adding LOCF for the new variables! Could you please update the section "2. Imputation by last-value-carried forward "LOCF" (`method` = 1)" of the documentation as well?

dschlaep, line 553: Nice job with adding LOCF capability to `generateMissingWeather()` for the new variables! Could you please update the section "2. Imputation by last-value-carried forward "LOCF" (`method` = 1)" of the documentation as well?

src/SW_Weather.c
dschlaep: `_read_weather_hist()` still uses `SW_Weather`, etc.

src/SW_Weather.c
```diff
@@ -781,6 +867,108 @@ void generateMissingWeather(
   }
 }
 
+/**
+ @brief Check weather through all years/days within simultation and make sure all input values are reasonable
+ after possible weather generation and scaling. If a value is to be found unreasonable, the function will execute a program crash.
+
+ @param weather Struct of type SW_WEATHER holding all relevant information pretaining to weather input data
+ */
+void checkAllWeather(SW_WEATHER *weather) {
```
dschlaep, line 916: This needs to be able to ignore missing weather values (at least for max/min temperature and ppt, and probably for all).

src/SW_Weather.c
```diff
@@ -781,6 +867,108 @@ void generateMissingWeather(
...
+
+    // Initialize any variables
+    TimeInt year, doy, numDaysInYear;
+    SW_WEATHER_HIST **weathHist = weather->allHist;
+
+    double dailyMinTemp, dailyMaxTemp;
+
+    // Check if minimum or maximum temperature, or precipitation flags are 0
+    if(!weather->has_temp2 || !weather->has_ppt) {
```
dschlaep, line 925: Note sure that `checkAllWeather()` is the best location to fail if `!weather->has_temp2 || !weather->has_ppt` -- we want to enforce that on the input files, i.e., maybe `SW_WTH_read()` or `readAllWeather()` seem places that more directly deal with input files

src/SW_Weather.c
```diff
@@ -945,7 +1138,12 @@ void SW_WTH_new_day(void) {
     /* get the daily weather from allHist */
     if (
       missing(w->allHist[yearIndex]->temp_avg[day]) ||
-      missing(w->allHist[yearIndex]->ppt[day])
+      missing(w->allHist[yearIndex]->ppt[day]) ||
+      missing(w->allHist[yearIndex]->cloudcov_daily[day]) ||
+      missing(w->allHist[yearIndex]->windspeed_daily[day]) ||
+      missing(w->allHist[yearIndex]->r_humidity_daily[day]) ||
+      (missing(w->allHist[yearIndex]->shortWaveRad[day]) && w->has_rsds) ||
```
dschlaep, line 1185: Could you please add a comment in the code why we treat radiation here differently, i.e., `(missing(w->allHist[yearIndex]->shortWaveRad[day]) && w->has_rsds)`?

We will add the special treatment for cloud cover with the next phase?

src/SW_Weather.c
```diff
@@ -1177,13 +1560,81 @@ void _read_weather_hist(
...
+                yearWeather->windspeed_daily[doy] = weathInput[windSpeedIndex];
+
+            } else if(weath->has_windComp) {
+                yearWeather->windspeed_daily[doy] = sqrt(squared(weathInput[windComp1Index]) +
+                                                    squared(weathInput[windComp2Index]));
+
+            }
+        }
+
+        if(!weath->use_relHumidityMonthly) {
```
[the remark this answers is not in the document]
dschlaep: `_read_weather_hist()` still does not quite implement calculations and priorities as we need them according to the google document:
- `vp` is currently not (but needs to be) calculated if `huss` is the only humidity input
- `r_humidity_daily` is currently not (but needs to be) calculated if `vp` is the only humidity input
- why is `if(!weath->has_vp) {...} else {...}` placed outside of `if(!weath->use_humidityMonthly)`? `_read_weather_hist()` handles only (and all) calculations from daily inputs

Additionally,
- the condition `weath->has_hurs2 && weath->has_temp2` does not work as intended because temperature values can still be missing even if `weath->has_temp2`; thus, `yearWeather->actualVaporPressure[doy]` needs to be set to missing if any of the temperature values are missing, otherwise, bogus calculations occur
- similarly, if `weath->has_huss`, then `yearWeather->r_humidity_daily[doy]` needs to be set to missing if `yearWeather->temp_avg[doy]` is missing, otherwise, bogus calculations occur

src/SW_Weather.c
```diff
@@ -1161,12 +1575,17 @@ void _read_weather_hist(
 
 	while (GetALine(f, inbuf)) {
 		lineno++;
-		x = sscanf(inbuf, "%d %f %f %f", &doy, &tmpmax, &tmpmin, &ppt);
-		if (x < 4) {
+		x = sscanf(inbuf, "%d %f %f %f %f %f %f %f %f %f %f %f %f %f %f",
```
dschlaep, line 1578: please address warning/error as seen on GHA (ubuntu-latest):
```
src/SW_Weather.c:1578:37: error: format ‘%d’ expects argument of type ‘int *’, but argument 3 has type ‘unsigned int *’ [-Werror=format=]
 1578 |                 x = sscanf(inbuf, "%d %f %f %f %f %f %f %f %f %f %f %f %f %f %f",
      |                                    ~^
      |                                     |
      |                                     int *
      |                                    %d
 1579 |                    &doy, &weathInput[0], &weathInput[1], &weathInput[2], &weathInput[3],
      |                    ~~~~              
      |                    |
      |                    unsigned int *
```

src/SW_Weather.c
```diff
@@ -502,29 +503,81 @@ void driestQtrSouthAdjMonYears(int month, int *adjustedYearZero, int *adjustedYe
  @param[in] use_weathergenerator_only A boolean; if `swFALSE`, code attempts to
    read weather files from disk.
  @param[in] weather_prefix File name of weather data without extension.
+ @param[in] use_cloudCoverMonthly A boolean; if `swTRUE`, function will interpolate
```
dschlaep, line 506: Please make documentation more precise for all three `use_*Monthly`
- `"read-in from disk"` is not correct for the scope of this function
- instead, `"... interpolate mean monthly values provided by \ref cloudcov to daily time series"`

src/SW_Weather.c
```diff
@@ -1124,11 +1575,28 @@ void SW_WTH_read(void) {
...
 */
 void _read_weather_hist(
   TimeInt year,
   SW_WEATHER_HIST *yearWeather,
-  char weather_prefix[]
+  char weather_prefix[],
+  unsigned int n_input_forcings,
+  unsigned int *dailyInputIndices,
+  Bool *dailyInputFlags,
+  Bool use_cloudCoverMonthly,
```
dschlaep, line 1597: I think we should get rid of the three monthly use flags from `_read_weather_hist()`
- this function is about daily values
- `SW_WTH_setup()` turns off daily use flags if monthly use flags are turned on -- making the information here redundant

src/SW_Weather.c
```diff
@@ -1177,13 +1656,103 @@ void _read_weather_hist(
...
+                relHum = e / es;
+                relHum = max(0., relHum);
+
+                yearWeather->r_humidity_daily[doy] = min(100., relHum);
+
+                // Check if a calculation of actual vapor pressure occurred with
+                // bad temperature values (SW_MISSING)
+                if(missing(yearWeather->temp_avg[doy])) {
+
+                    yearWeather->r_humidity_daily[doy] = SW_MISSING;
```
dschlaep, line 1713: Why calculate relative humidity from specific humidity if we are going to override it anyhow? Why not an if-else?
```
if(!missing(yearWeather->temp_avg[doy])) {
  // do the calculation
} else {
  // set to missing
}
```

src/SW_Weather.c
```diff
@@ -1177,13 +1656,103 @@ void _read_weather_hist(
...
+                                actualVaporPressure2(weathInput[dailyInputIndices[REL_HUMID_MAX]],
+                                                     weathInput[dailyInputIndices[REL_HUMID_MIN]],
+                                                     weathInput[dailyInputIndices[TEMP_MAX]],
+                                                     weathInput[dailyInputIndices[TEMP_MIN]]);
+
+                // Check if a calculation of relative humidity occurred with
+                // bad temperature values (SW_MISSING)
+                if(missing(yearWeather->temp_max[doy]) || missing(yearWeather->temp_min[doy])) {
+
+                    yearWeather->r_humidity_daily[doy] = SW_MISSING;
```
dschlaep, line 1738: Why set relative humidity to missing? This should be actual vapor pressure

src/SW_Weather.c
```diff
@@ -1177,13 +1656,103 @@ void _read_weather_hist(
...
+            } else if(hasMaxMinTemp && hasMaxMinRelHumid) {
+                yearWeather->actualVaporPressure[doy] =
+                                actualVaporPressure2(weathInput[dailyInputIndices[REL_HUMID_MAX]],
+                                                     weathInput[dailyInputIndices[REL_HUMID_MIN]],
+                                                     weathInput[dailyInputIndices[TEMP_MAX]],
+                                                     weathInput[dailyInputIndices[TEMP_MIN]]);
+
+                // Check if a calculation of relative humidity occurred with
+                // bad temperature values (SW_MISSING)
+                if(missing(yearWeather->temp_max[doy]) || missing(yearWeather->temp_min[doy])) {
```
dschlaep, line 1736: same comment as for specific humidity: why not an `if-else`?

src/SW_Weather.c
```diff
@@ -1177,13 +1656,103 @@ void _read_weather_hist(
...
+
+                    yearWeather->r_humidity_daily[doy] = SW_MISSING;
+                }
+
+            } else if(dailyInputFlags[REL_HUMID] || dailyInputFlags[SPEC_HUMID]) {
+                yearWeather->actualVaporPressure[doy] =
```
dschlaep, line 1742: `temp_avg` may be missing, so we need to protect this here as well:
```
if (!missing(yearWeather->temp_avg[doy])) {
  yearWeather->actualVaporPressure[doy] =
               actualVaporPressure1(yearWeather->r_humidity_daily[doy],
                                    yearWeather->temp_avg[doy]);
} else {
yearWeather->actualVaporPressure[doy] = SW_MISSING;
}
```

src/SW_Weather.c
```diff
@@ -1177,13 +1656,103 @@ void _read_weather_hist(
...
+                             actualVaporPressure1(yearWeather->r_humidity_daily[doy],
+                                                  yearWeather->temp_avg[doy]);
+            }
+
+            // Check if a better calculation of relative humidity is available
+            // (using dewpoint temperature or the daily actual vapor pressure input value)
+            if(dailyInputFlags[ACTUAL_VP] || dailyInputFlags[TEMP_DEWPOINT]) {
```
dschlaep, line 1749: Calculate `yearWeather->r_humidity_daily[doy]` only from vapor pressure or dewpoint temperature if not otherwise provided as input, i.e., if missing

src/SW_Weather.c
```diff
@@ -1177,13 +1656,103 @@ void _read_weather_hist(
...
+
+                svpVal = svp(yearWeather->temp_avg[doy], &tempSlope);
```
dschlaep, line 1751: safeguard calculation of relative humidity from vapor pressure against missing values in temperature or vapor pressure

src/SW_Weather.c
N1ckP3rsl3y: I meant to mention this to you. I think it would be best to remove cloud cover and wind speed monthly flags. I think it would be a good idea, unless argued otherwise, to keep `use_humidityMonthly`. I think this because it's more efficient to skip the entire code block dealing with humidity values. I.e., instead of having all daily flags be turned off and check these flags ~11 times a day instead of one (`use_humidityMonthly).

dschlaep: Good point considering how many logical conditions would be tested daily -- I didn't think about that aspect! I still think we should remove all three monthly use flags to clean up the interface; what do think about creating a local variable `use_humidityDaily` that is calculated from the daily flags and which we can use instead of `!use_humidityMonthly`?

N1ckP3rsl3y: I think that would be a good idea since it would, like you stated, make the interface easier to deal with. I think your idea is the best. I just need to see if keeping it static would be a good idea. I don't believe so since the initial testing and setting of the value would be difficult due to the fact that we cannot use SW_MISSING, since it's of type Bool. It's possible to make it static but may require more than it's worth. The comparison of number of checks in 31 years would be ~68,076 > ~186 > ~6 (completely getting rid of monthly humidity flag, using your idea as non-static, using your idea as a static variable, respectively).

Though, am I mistaken, or if we are to solely calculate if `use_humidityMonthly` from other flags, is there not a chance a user can turn off the monthly flags and have none of the subsequent daily flags set, so the program assumes that the monthly flag is in use? This should just result in a crash within `SW_WTH_new_day()`, correct?

tests/gtests/test_SW_Weather.cc
```diff
@@ -686,5 +698,315 @@ namespace {
...
+
+             // Based on the max/min of relative humidity, test if actual vapor pressure
+             // was calculated reasonably
+         EXPECT_NEAR(result, expectedResult, tol6);
+
+         // Make sure calculations and set input values are within reasonable range
+         checkAllWeather(w);
+
+         // Reset directory to "data_weather"
+         strcpy(w->name_prefix, "Input/data_weather/weath");
```
dschlaep, line 794: This test (and others that modify global state) need to fully reset all global states (instead of partially with `strcpy(w->name_prefix, "Input/data_weather/weath");`
```
    // Reset to previous global state
    Reset_SOILWAT2_after_UnitTest();
```

tests/gtests/test_SW_Weather.cc
```diff
@@ -686,5 +698,315 @@ namespace {
...
+             w->n_input_forcings,
+             w->dailyInputIndices,
+             w->dailyInputFlags
+         );
+
+         result = w->allHist[yearIndex]->r_humidity_daily[0];
+
+             // Get expected average from Input/data_weather_gridmet/weath.1980 day 1
+             // hursmax_pct and hursmin_pct
+         expectedResult = (74.17 + 31.42) / 2.;
```
dschlaep, line 774: here and the other similar expects: it would be great to demonstrate that we have values based on daily inputs and not based on the mean monthly interpolated inputs, e.g.,

```
// Expect that daily relative humidity is derived from hursmax_pct and hursmin_pct
// (and is not interpolated from mean monthly values)
EXPECT_NEAR(
  w->allHist[yearIndex]->r_humidity_daily[midJanDay],
  (74.17 + 31.42) / 2.,
  tol6
);
EXPECT_NE(
  w->allHist[yearIndex]->r_humidity_daily[midJanDay],
  SW_Sky.r_humidity[0],
  tol6
);
```

src/SW_Weather.c
```diff
@@ -1662,30 +1697,27 @@ void _read_weather_hist(
...
-            if(dailyInputFlags[WIND_SPEED]) {
-                yearWeather->windspeed_daily[doy] = weathInput[dailyInputIndices[WIND_SPEED]];
-            } else if(dailyInputFlags[WIND_EAST] && dailyInputFlags[WIND_NORTH]) {
-                yearWeather->windspeed_daily[doy] = sqrt(squared(weathInput[dailyInputIndices[WIND_EAST]]) +
-                                                         squared(weathInput[dailyInputIndices[WIND_NORTH]]));
-            }
+        if(dailyInputFlags[WIND_SPEED]) {
+            yearWeather->windspeed_daily[doy] = weathInput[dailyInputIndices[WIND_SPEED]];
+
+        } else if(dailyInputFlags[WIND_EAST] && dailyInputFlags[WIND_NORTH]) {
```
dschlaep, line 1713: Considering `_read_weather_hist()` a bit further: we need to capture even more cases of missing values and propagate them through to `allHist`. If we just go ahead and calculate with SW_MISSING (which is just 999) then we produce a value that is not SW_MISSING), e.g., `500.5 = (2 + SW_MISSING) / 2)` which is a nonsense value. In order for `SW_WTH_new_day()` to actually be able to fail if a value is missing, we need to propagate all missing input values. Thus, the following cases must additionally propagate SW_MISSING (such as for instance `yearWeather->temp_avg[doy] = ...` is doing):
- wind speed calculated from east/north wind components
- relative humidity calculated from min/max relative humidity
- relative humidity calculated from specific humidity
- vapor pressure calculated from dewpoint temperature
- vapor pressure calculated from min/max relative humidity
- vapor pressure calculated from relative humidity

Cases where we just copy input values into our daily data structure don't need to be handled separately, e.g., relative humidity copied from input relative humidity (case `dailyInputIndices[REL_HUMID]`).

dschlaep (commented):
Nice progress! A few more comments and realizing that `_read_weather_hist()` needs to propagate all missing values from inputs to our daily variables.

dschlaep (commented):
Nice job!

Our checks are currently failing on [GHA for windows/cygwin](https://github.com/DrylandEcology/SOILWAT2/actions/runs/4022536419/jobs/6912378488). Maybe some of our new tests still cause side effects?
```
[----------] 1 test from VegEstabTest
[ RUN      ] VegEstabTest.SimulateWithVegEstab
@ generic.c LogErrormake: *** [makefile:344: test_run] Error 255
Error: Process completed with exit code 2.
```

It would also be great to add a test that checks that `generateMissingWeather()` does indeed successfully fill in missing values in the new daily variables using the `LOCF` method.

dschlaep:
Running our tests in an infinite loop with random order (until one crashes) may point to the same issue that causes our `build_Windows_Cygwin` to crash, @N1ckP3rsl3y ?

For instance, I get

// commit e44035225c6e951331712a044abf1355f83c616c
// with Apple clang version 14.0.0
`make clean test`

// Run tests in an infinite loop with random order (until one crashes)
`bin/sw_test --gtest_shuffle --gtest_repeat=-1`


// For instance, here a few crashes
`bin/sw_test --gtest_shuffle --gtest_random_seed=69894`

```
[----------] 9 tests from WaterBalanceTest
[ RUN      ] WaterBalanceTest.WithSoilTemperature
@ generic.c LogError
```

`bin/sw_test --gtest_shuffle --gtest_random_seed=74898`

```
[----------] 9 tests from WaterBalanceTest
[ RUN      ] WaterBalanceTest.WithPondedWaterRunonRunoff
@ generic.c LogError
```

`bin/sw_test --gtest_shuffle --gtest_random_seed=70537`

```
[----------] 2 tests from SWSiteTest
[ RUN      ] SWSiteTest.SoilTranspirationRegions
[       OK ] SWSiteTest.SoilTranspirationRegions (7 ms)
[ RUN      ] SWSiteTest.SoilDensity
@ generic.c LogError
```

or

```
[----------] 9 tests from WaterBalanceTest
[ RUN      ] WaterBalanceTest.Example1
@ generic.c LogError
```

dschlaep (commented):
great job!

dschlaep:
- [ ] Update NEWS
- [ ] Make sure it works for rSOILWAT2
- [ ] Propagate latest changes to STEPWAT2 branch 'read_more_daily_input'

Issue #346, opened 2023-02-20
Reentrancy and thread-safety


dschlaep:
Convert SOILWAT2 to be reentrant and thread-safe

For instance:
- global variables, e.g., `SW_Weather`, `SW_Site` 
    - replace with variables local to `main()` (SW_Main.c, sw_maintest.cc, etc.)
    - consider a new "meta-"/"super-" struct that holds all these currently global variables
    - pass by reference as function arguments

- functions that return a pointer to static data, e.g., `DirName()`
    - replace with caller-provided storage (see `str_toupper()` example in [programming-writing-reentrant-threadsafe-code](https://www.ibm.com/docs/en/aix/7.2?topic=programming-writing-reentrant-threadsafe-code)

- module-level local variables
    - replace with function-level local variable, e.g., `MyFilename`
    - move to a new/existing struct and create instance in `main()` (see section "global variables"), e.g., `memoized_G_o`

- "static const" variables should not be a problem per se
- "static" functions should not be a problem per se

this is not an exhaustive list

PR #349, opened 2023-04-03
Reentrancy & Threadsafety

dschlaep:
- [ ] #346

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # NEWS
 
-# SOILWAT2 v7.0.0-9000
+# SOILWAT2 v7.0.0
 * This version produces nearly identical simulation output
   as the previous release under default values for the new inputs.
   Small deviations arise due to a fix in the handling of soil moisture values
@@ -32,6 +32,26 @@
       by `finalizeAllWeather()` via `SW_WTH_finalize_all_weather()`;
       this must occur before the simulation is "initialized"
       by `SW_CTL_init_run()`.
+    * Imputation of missing daily weather values by `generateMissingWeather()`
+      can be done either by last-observation-carried-forward `"LOCF"`
+      (which handles all daily weather variables) or by the Markov weather
+      generator (which handles only temperature and precipitation).
+
+* Daily weather inputs, in addition to the previous variables
+  maximum air temperature, minimum air temperature, and precipitation amount,
+  can now process the following variables (issue #341; @dschlaep, @N1ckP3rsl3y):
+    * Cloud cover (can be replaced by shortwave radiation)
+    * Wind speed (can be replaced by wind components)
+    * Wind speed eastward component (optional)
+    * Wind speed northward component (optional)
+    * Relative humidity (can be replaced by max/min humidity, specific humidity,
+      dew point temperature, or vapor pressure)
+    * Maximum relative humidity (optional)
+    * Minimum relative humidity (optional)
+    * Specific humidity (optional)
+    * Dew point temperature (optional)
+    * Actual vapor pressure (optional)
+    * Downward surface shortwave radiation (optional)
 
 * SOILWAT2 gains the ability to calculate long-term climate summaries
   (issue #317; @N1ckP3rsl3y, @dschlaep).
@@ -71,6 +91,13 @@
 
 
 ## Changes to inputs
+* New inputs via `"weathsetup.in"` determine whether monthly or daily inputs
+  for cloud cover, relative humidity, and wind speed are utilized;
+  describe which daily weather variables are contained in the weather input
+  files `weath.YYYY`; and describe units of (optional) input shortwave
+  radiation.
+* New (optional) variables (columns) in weather input files `weath.YYYY` that
+  are described via `"weathsetup.in"`.
 * New inputs via `"siteparam.in"` select a soil water release curve `swrc_name`
   and determine parameter source `has_swrcp`, i.e.,
   either estimated via selected pedotransfer function `ptf_name` or
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # NEWS
 
+# SOILWAT2 v7.1.0-9000
+
 # SOILWAT2 v7.0.0
 * This version produces nearly identical simulation output
   as the previous release under default values for the new inputs.
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,10 @@
 # NEWS
 
 # SOILWAT2 v7.1.0-9000
+* Bugfixes
+    * Fix an error where a pointer was freed even though it was not allocated
+      (issue #356; @dschlaep).
+    * Fix memory leak in test of `SW_VPD_construct()` (issue #205; @dschlaep).
 
 # SOILWAT2 v7.0.0
 * This version produces nearly identical simulation output
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,10 +1,38 @@
 # NEWS
 
 # SOILWAT2 v7.1.0-9000
+* Simulation output remains the same as the previous version.
+
+* Prepare for SOILWAT2 to become thread-safe and reentrant (#346; @N1ckP3rsl3y)
+    * Definition clarifications
+        * Thread-safe - Multiple threads (a future SOILWAT2 development)
+          will not influence each other unintentionally. Here, we implemented
+          structures to enable thread-local storage, i.e., each thread operates
+          on local data structures or with static data.
+        * Reentrant - The ability to correctly execute any part of the program
+          by multiple threads simultaneously but independently from others.
+    * All non-static variables are replaced by local variables;
+      functions gained arguments to pass local variables by reference
+      (replacing global variables).
+    * New main abstract types
+        * SW_ALL - Contains the existing structures that are relevant
+          for the simulation itself, e.g., SW_MODEL, SW_SOILWAT, SW_WEATHER.
+        * PATH_INFO - Holds information about location of input and output data,
+          e.g., directories, file paths to input data.
+        * SW_OUTPUT_POINTERS - Points to requested output subroutines.
+        * LOG_INFO - Manages information for logging warnings and errors.
+
 * Bugfixes
     * Fix an error where a pointer was freed even though it was not allocated
       (issue #356; @dschlaep).
     * Fix memory leak in test of `SW_VPD_construct()` (issue #205; @dschlaep).
+
+## Changes to inputs
+* The output separator `OUTSEP` has been unused since the switch (`v4.0.0`)
+  from writing free-form text files to `"csv"` spreadsheets
+  (where the separator is fixed to a comma); the occurrence of
+  `OUTSEP` in `"outsetup.in"` is now deprecated and triggers a warning.
+
 
 # SOILWAT2 v7.0.0
 * This version produces nearly identical simulation output
--- a/include/SW_Files.h
+++ b/include/SW_Files.h
@@ -52,7 +52,7 @@ typedef enum {
 /* =================================================== */
 /*             Global Function Declarations            */
 /* --------------------------------------------------- */
-void SW_F_read(LOG_INFO* LogInfo, PATH_INFO* PathInfo);
+void SW_F_read(PATH_INFO* PathInfo, LOG_INFO* LogInfo);
 void SW_F_construct(char *InFiles[], const char *firstfile, char _ProjDir[]);
 void SW_F_deconstruct(char *InFiles[]);
 void SW_CSV_F_INIT(const char *s, LOG_INFO* LogInfo);
--- a/include/SW_Files.h
+++ b/include/SW_Files.h
@@ -53,7 +53,8 @@ typedef enum {
 /*             Global Function Declarations            */
 /* --------------------------------------------------- */
 void SW_F_read(PATH_INFO* PathInfo, LOG_INFO* LogInfo);
-void SW_F_construct(char *InFiles[], const char *firstfile, char _ProjDir[]);
+void SW_F_construct(char *InFiles[], const char *firstfile, char _ProjDir[],
+					LOG_INFO *LogInfo);
 void SW_F_deconstruct(char *InFiles[]);
 void SW_CSV_F_INIT(const char *s, LOG_INFO* LogInfo);
 
--- a/include/SW_Flow_lib.h
+++ b/include/SW_Flow_lib.h
@@ -65,12 +65,12 @@ void EsT_partitioning(double *fbse, double *fbst, double blivelai, double lai_pa
 
 void pot_soil_evap(SW_SITE *SW_Site, unsigned int nelyrs, double totagb,
   double fbse, double petday, double shift, double shape, double inflec,
-  double range, double swc[], double Es_param_limit, LOG_INFO* LogInfo,
-  double *bserate);
+  double range, double swc[], double Es_param_limit, double *bserate,
+  LOG_INFO* LogInfo);
 
-void pot_soil_evap_bs(double *bserate, SW_SITE *SW_Site, LOG_INFO* LogInfo,
+void pot_soil_evap_bs(double *bserate, SW_SITE *SW_Site,
 	unsigned int nelyrs, double petday, double shift, double shape,
-	double inflec, double range, double swc[]);
+	double inflec, double range, double swc[], LOG_INFO* LogInfo);
 
 void pot_transp(double *bstrate, double swpavg, double biolive, double biodead, double fbst, double petday, double swp_shift, double swp_shape, double swp_inflec,
 		double swp_range, double shade_scale, double shade_deadmax, double shade_xinflex, double shade_slope, double shade_yinflex, double shade_range, double co2_wue_multiplier);
@@ -146,10 +146,10 @@ void soil_temperature(ST_RGR_VALUES* SW_StRegValues,
 					  double H_gt,
 					  TimeInt year,
 					  TimeInt doy,
-					  LOG_INFO* LogInfo,
 					  double maxLyrTemperature[],
 					  double minLyrTemperature[],
-					  Bool *ptr_stError);
+					  Bool *ptr_stError,
+					  LOG_INFO* LogInfo);
 
 void lyrTemp_to_lyrSoil_temperature(double cor[MAX_ST_RGR][MAX_LAYERS + 1],
   unsigned int nlyrTemp, double depth_Temp[], double avgLyrTempR[], unsigned int nlyrSoil,
@@ -185,15 +185,15 @@ void SW_ST_setup_run(
 	double deltaX,
 	double theMaxDepth,
 	unsigned int nRgr,
-	LOG_INFO* LogInfo,
 	double *surfaceAvg,
-	double* lyrFrozen
+	double* lyrFrozen,
+	LOG_INFO* LogInfo
 );
 
 void soil_temperature_setup(ST_RGR_VALUES* SW_StRegValues, double bDensity[],
 	double width[], double oldavgLyrTemp[], double sTconst, unsigned int nlyrs,
 	double fc[], double wp[], double deltaX, double theMaxDepth,
-	unsigned int nRgr, LOG_INFO* LogInfo, Bool *ptr_stError, Bool *soil_temp_init);
+	unsigned int nRgr, Bool *ptr_stError, Bool *soil_temp_init, LOG_INFO* LogInfo);
 
 void set_frozen_unfrozen(unsigned int nlyrs, double avgLyrTemp[], double swc[],
 						 double swc_sat[], double width[], double lyrFrozen[]);
--- a/include/SW_Flow_lib.h
+++ b/include/SW_Flow_lib.h
@@ -57,20 +57,20 @@ void infiltrate_water_high(double swc[], double drain[], double *drainout, doubl
 	int nlyrs, double swcfc[], double swcsat[], double impermeability[],
 	double *standingWater, double lyrFrozen[]);
 
-void transp_weighted_avg(double *swp_avg, unsigned int n_tr_rgns, LyrIndex n_layers,
-						 unsigned int tr_regions[], double tr_coeff[], double swc[],
-						 SW_LAYER_INFO** lyr, LOG_INFO* LogInfo);
+void transp_weighted_avg(double *swp_avg, SW_SITE *SW_Site,
+	unsigned int n_tr_rgns, LyrIndex n_layers, unsigned int tr_regions[],
+	double swc[], int VegType, LOG_INFO* LogInfo);
 
 void EsT_partitioning(double *fbse, double *fbst, double blivelai, double lai_param);
 
-void pot_soil_evap(unsigned int nelyrs, double ecoeff[], double totagb,
-  double fbse, double petday, double shift, double shape, double inflec, double range,
-  double width[], double swc[], double Es_param_limit, SW_LAYER_INFO** lyr,
-  LOG_INFO* LogInfo, double *bserate);
+void pot_soil_evap(SW_SITE *SW_Site, unsigned int nelyrs, double totagb,
+  double fbse, double petday, double shift, double shape, double inflec,
+  double range, double swc[], double Es_param_limit, LOG_INFO* LogInfo,
+  double *bserate);
 
-void pot_soil_evap_bs(double *bserate, SW_LAYER_INFO** lyr, LOG_INFO* LogInfo,
-	unsigned int nelyrs, double ecoeff[], double petday, double shift,
-	double shape, double inflec, double range, double width[], double swc[]);
+void pot_soil_evap_bs(double *bserate, SW_SITE *SW_Site, LOG_INFO* LogInfo,
+	unsigned int nelyrs, double petday, double shift, double shape,
+	double inflec, double range, double swc[]);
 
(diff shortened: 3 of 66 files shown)


src/SW_Flow_lib.c
```diff
@@ -1712,17 +1742,18 @@ Based on equations from Eitzinger 2000. @cite Eitzinger2000
...
-temp = oldavgLyrTemp[0] + avgLyrTemp[0] + shParam + nlyrs + vwc[0] + bDensity[0];
-temp += temp;
-// end avoid compiler warnings
+	//	double deltaTemp, Cis, sFusionPool[nlyrs], sFusionPool_actual[nlyrs];
+	// To avoid compiler warnings "warning: parameter set but not used"
+	double temp;
+	temp = oldavgLyrTemp[0] + avgLyrTemp[0] + shParam + nlyrs + vwc[0] + bDensity[0];
+	temp += temp;
+	// end avoid compiler warnings
+	ST_RGR_VALUES dummyInstance;
```
dschlaep, line 1756: Why convert the global object `ST_RGR_VALUES *st = &stValues;`  into a local copy `ST_RGR_VALUES dummyInstance;`?

Why not upgrade to a function argument `ST_RGR_VALUES* SW_StRegValues` so that it can be modified (as done elsewhere)?

dschlaep (commented):
A few small comments. Thanks!

src/SW_Flow_lib.c
dschlaep: `adjust_Tsoil_by_freezing_and_thawing()`

src/SW_Site.c
```diff
@@ -1296,20 +1300,22 @@ void add_deepdrain_layer(void) {
...
+	SW_Site->n_layers++;
 
-	v->lyr = (!v->lyr) /* if not yet defined */
-		? (SW_LAYER_INFO **) Mem_Calloc(v->n_layers, sizeof(SW_LAYER_INFO *), "_newlayer()") /* malloc() it  */
-		: (SW_LAYER_INFO **) Mem_ReAlloc(v->lyr, sizeof(SW_LAYER_INFO *) * (v->n_layers)); /* else realloc() */
+	SW_Site->lyr = (!SW_Site->lyr) /* if not yet defined */
+		? (SW_LAYER_INFO **) Mem_Calloc(SW_Site->n_layers, sizeof(SW_LAYER_INFO *), "_newlayer()") /* malloc() it  */
+		: (SW_LAYER_INFO **) Mem_ReAlloc(SW_Site->lyr, sizeof(SW_LAYER_INFO *) * (SW_Site->n_layers)); /* else realloc() */
 
-	v->lyr[v->n_layers - 1] = (SW_LAYER_INFO *) Mem_Calloc(1, sizeof(SW_LAYER_INFO), "_newlayer()");
```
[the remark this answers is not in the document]
N1ckP3rsl3y: The variable "lyr" is still base0. At the beginning of the function, the value is stored in a variable called `n_layers` and then `SW_Site->layers` is incremented. So when the function needs the original or incremented value, it will use `SW_Site->layers` or `n_layers`, respectively. This was done to reduce the instances of decrementing by one in multiple spots.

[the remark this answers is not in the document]
dschlaep: make sense, nice implementation! My bad for not realizing the smarter way that you implemented

tests/gtests/sw_testhelpers.cc
```diff
@@ -15,18 +15,22 @@
...
 #include "include/SW_SoilWater.h"
 #include "include/SW_Weather.h"
 #include "include/SW_Control.h"
-#include "include/SW_Main_lib.h" // for `sw_check_log()`
+#include "include/SW_Files.h"
+#include "include/myMemory.h"
 
 #include "tests/gtests/sw_testhelpers.h"
 
+SW_OUTPUT_POINTERS SW_OutputPtrs;
```
dschlaep, line 29: Why are some of the gtest global variables defined in `sw_maintest.cc` and others in `sw_testhelpers.cc`?

include/SW_Files.h
```diff
@@ -51,25 +49,15 @@ typedef enum {
...
 /*             Global Function Declarations            */
 /* --------------------------------------------------- */
-void SW_F_read(const char *s);
-char *SW_F_name(SW_FileIndex i);
-void SW_F_construct(const char *firstfile);
-void SW_F_deconstruct(void);
-void SW_WeatherPrefix(char prefix[]);
-void SW_CSV_F_INIT(const char *s);
+void SW_F_read(LOG_INFO* LogInfo, PATH_INFO* PathInfo);
+char *SW_F_name(SW_FileIndex i, char *InFiles[]);
```
dschlaep, line 56: Couldn't we just get rid of `SW_F_name()` and replace with `InFiles[i]`?

include/SW_Flow_lib.h
```diff
@@ -64,18 +59,18 @@ void infiltrate_water_high(double swc[], double drain[], double *drainout, doubl
...
-						 SW_LAYER_INFO** lyr);
+						 SW_LAYER_INFO** lyr, LOG_INFO* LogInfo);
 
 void EsT_partitioning(double *fbse, double *fbst, double blivelai, double lai_param);
 
 void pot_soil_evap(unsigned int nelyrs, double ecoeff[], double totagb,
   double fbse, double petday, double shift, double shape, double inflec, double range,
   double width[], double swc[], double Es_param_limit, SW_LAYER_INFO** lyr,
-  double *bserate);
+  LOG_INFO* LogInfo, double *bserate);
```
dschlaep, line 69: Is there a reason why the argument `LOG_INFO* LogInfo` does show up in virtually any position of the argument list across SOILWAT2 functions?

I feel it could help if it is treated differently than other arguments since it doesn't do anything with how the simulation works and is only about the user interface relaying messages, warnings, and errors. For instance, having it consistently as the last argument would seem to make the ordering of it much simpler (except for `LogError()` for which a first position does make most sense).

include/SW_datastructs.h
```diff
@@ -689,6 +739,22 @@ typedef struct {
 
 } SW_SOILWAT;
 
+typedef struct {
+	FILE *logfp; /* REQUIRED; defined in SW_Main_lib.c */
```
dschlaep, line 743: The comment `/* REQUIRED; defined in SW_Main_lib.c */` doesn't apply anymore for both `*logfp` and `logged`

include/SW_Files.h
```diff
@@ -51,25 +49,15 @@ typedef enum {
...
+void SW_F_construct(char *InFiles[], const char *firstfile, char _ProjDir[]);
+void SW_F_deconstruct(char *InFiles[]);
+void SW_WeatherPrefix(char prefix[], char weather_prefix[]);
```
dschlaep, line 59: Same for `SW_WeatherPrefix()`: couldn't we just inline the one line of code for the one time it is used?

dschlaep (commented):
Lots of work! A few small comments plus thoughts about the placement of `LogInfo` argument -- let's discuss in our next meeting. Thanks!

tests/gtests/sw_testhelpers.cc
```diff
@@ -80,7 +74,8 @@ void Reset_SOILWAT2_after_UnitTest(void) {
 
 
 
-void create_test_soillayers(unsigned int nlayers, char *InFiles[]) {
+void create_test_soillayers(unsigned int nlayers, char *InFiles[],
```
dschlaep, line 77: `create_test_soillayers()` also utilizes "LogInfo" and needs this is an argument

tests/gtests/sw_testhelpers.h
```diff
@@ -8,6 +8,8 @@
 static const double
   tol0 = 1e-0, tol1 = 1e-1, tol2 = 1e-2, tol3 = 1e-3, tol6 = 1e-6, tol9 = 1e-9;
 
+extern const char * masterfile_test = "files.in"; // relative to 'dir_test'
```
dschlaep, line 11: `*masterfile_test` is currently neither defined nor declared -- so the extern doesn't work here. Maybe remove the "extern" keyword

include/SW_datastructs.h
```diff
@@ -0,0 +1,1055 @@
...
+
+		/** \brief A 2-dim array of pointers to output arrays.
+
+	\note This should be initialized to NULL because they are defined globally
+		and thus have `static storage duration`.
+
+	The variable p_OUT used by rSOILWAT2 for output and by STEPWAT2 for
+	mean aggregation.
+	*/
+	RealD *p_OUT[SW_OUTNKEYS][SW_OUTNPERIODS];
```
dschlaep, line 964: Elements previously in "SW_Output_outarray" should be wrapped by `if defined(SW_OUTARRAY) ...` because the entire file "SW_Output_outarray.c" was only compiled by rSOILWAT2 or STEPWAT2 but not SOILWAT2 (see makefile and Makevars).

This applies for instance to `*p_OUT`, `nrow_OUT`, etc.

tests/gtests/sw_maintest.cc
```diff
@@ -31,6 +31,7 @@
    of the SOILWAT2 repository
 */
 const char * dir_test = "./tests/example";
+const char * masterfile_test = "files.in"; // relative to 'dir_test'
```
dschlaep, line 34: We have now `const char * masterfile_test = "files.in";` in "sw_maintest.cc" and `const char * filemaster_test = "files.in";` in "sw_testhelpers.cc" -- could we get rid of one? E.g., move it to "sw_testhelpers.h" and use in both places? And rename to something like `const char * filemain_test = "files.in";` when we are at it?

tests/gtests/test_SW_VegProd.cc
```diff
@@ -157,11 +157,19 @@ namespace {
 
   // Test summing values across vegetation types
   TEST(VegTest, Summing) {
```
dschlaep, line 159: In "VegTest, Summing" this now uses the test globals `SW_All.Site.transp_coeff`. If we do that then we need to appropriately reset the globals by calling `Reset_SOILWAT2_after_UnitTest();` at the end of the test.

src/SW_Site.c
```diff
@@ -2466,29 +2435,6 @@ void SW_SIT_init_run(SW_VEGPROD* SW_VegProd, SW_SITE* SW_Site,
 	}
 }
 
-/**
-@brief For multiple runs with the shared library, the need to remove the allocated
-			soil layers arises to avoid leaks. (rjm 2013)
-
-@param[in,out] SW_Site Struct of type SW_SITE describing the simulated site
-*/
-void SW_SIT_clear_layers(SW_SITE* SW_Site) {
```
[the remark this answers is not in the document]
N1ckP3rsl3y: The only scenario where we would need to reset `SW_Site->n_layers` would be in tests. However, this is done when the program resets. More specifically, in the function `SW_SIT_init_counts()`.

include/SW_Site.h
```diff
@@ -230,12 +230,11 @@ void PTF_RawlsBrakensiek1985(
 RealD calculate_soilBulkDensity(RealD matricDensity, RealD fractionGravel);
 RealD calculate_soilMatricDensity(RealD bulkDensity, RealD fractionGravel,
 								  LOG_INFO* LogInfo);
-LyrIndex nlayers_bsevap(SW_LAYER_INFO** lyr, LyrIndex n_layers);
+LyrIndex nlayers_bsevap(RealD *evap_coeff, LyrIndex n_layers);
 void nlayers_vegroots(LyrIndex n_layers, LyrIndex n_transp_lyrs[],
-					  SW_LAYER_INFO** lyr);
+					  RealD transp_coeff[][MAX_LAYERS + 1]);
```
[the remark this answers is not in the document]
N1ckP3rsl3y: Would it be possible to elaborate a bit more on what you mean by a new define? I understand the concern for the +1 but I don't think I understand what you're proposing.

src/SW_SoilWater.c
```diff
@@ -343,7 +343,7 @@ void SW_WaterBalance_Checks(SW_ALL* sw, LOG_INFO* LogInfo)
   RealD
     Etotal, Etotalsurf, Etotalint, Eponded, Elitter, Esnow, Esoil = 0., Eveg = 0.,
     Ttotal = 0., Ttotalj[MAX_LAYERS],
-    percolationIn[MAX_LAYERS + 1], percolationOut[MAX_LAYERS + 1],
+    percolationIn[MAX_LAYERS], percolationOut[MAX_LAYERS],
```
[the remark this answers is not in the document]
N1ckP3rsl3y: I believe this was just a mistake on my part and didn't mean to actually do it. It has been fixed

NEWS.md
```diff
@@ -1,6 +1,6 @@
 # NEWS
 
-# SOILWAT2 v7.0.0-9000
+# SOILWAT2 v7.0.0
```
dschlaep, line 3: Why did commit 0bcd6b205b80455be030fea63a44fd2fb84376d0 update from branch "release/devel_v7.0.0" (which no longer existed on origin at that time) instead from latest development branch "release/devel_v7.1.0" as requested? Please update

include/SW_Flow_lib.h
```diff
@@ -170,30 +170,21 @@ void SW_ST_init_run(ST_RGR_VALUES* StRegValues);
 
 void SW_ST_setup_run(
```
dschlaep, line 171: Using `*SW_Site` as argument for `SW_ST_setup_run()` is much nicer than all the individual parameters! Thanks to getting rid of the soil layer info struct!

tests/gtests/sw_testhelpers.h
```diff
@@ -18,14 +22,54 @@ static const double
 #undef missing
 #define missing(x)  ( EQ( fabs( (x) ), SW_MISSING ) || !std::isfinite( (x) ) )
 
-extern SW_ALL SW_All;
-extern SW_OUTPUT_POINTERS SW_OutputPtrs;
-extern LOG_INFO LogInfo;
-extern PATH_INFO PathInfo;
+
+class AllTest : public::testing::Test {
```
dschlaep, line 26: test fixtures are the way to go! It also appears that the tests are running slightly faster?

tests/gtests/test_SW_Carbon.cc
```diff
@@ -40,34 +40,29 @@
...
-
-    // Reset to previous global state
-    Reset_SOILWAT2_after_UnitTest();
   }
 
 
   // Test reading yearly CO2 data from disk file
-  TEST(CarbonTest, ReadInputFile) {
-    TimeInt year;
+  TEST_F(AllTest, ReadCarbonInputFile) {
```
dschlaep, line 56: Nice change from test name "ReadInputFile" to "ReadCarbonInputFile"! Since every test is now based on "AllTest" fixture -- we loose the name of the test suit. Please add such information as done here throughout

tests/gtests/test_SW_Flow_Lib_PET.cc
```diff
@@ -589,7 +589,7 @@ namespace
   //   Comparison against examples by Duffie & Beckman 2013 are expected to
   //   deviate in value, but show similar patterns, because equations for
   //   (i) sun-earth distance equation and (ii) solar declination differ
-  TEST(SW2SolarRadiationTest, extraterrestrial)
+  TEST_F(AllTest, extraterrestrial)
```
dschlaep, line 592: Please change `TEST_F(AllTest, extraterrestrial)` to something like `TEST_F(AllTest, SolarRadiationExtraterrestrial)`

tests/gtests/test_SW_Markov.cc
```diff
@@ -57,14 +57,12 @@ namespace {
...
 
-    // Reset to previous global state
-    // Reset_SOILWAT2_after_UnitTest();
     SW_MKV_deconstruct(&SW_All.Markov);
   }
 
 
   // Check seeding of RNG for weather generator
-  TEST(WGTest, Seeding) {
+  TEST_F(AllTest, Seeding) {
```
dschlaep, line 65: Please change `TEST_F(AllTest, Seeding)` to something like `TEST_F(AllTest, WeatherGeneratorRNGSeeding)`

tests/gtests/test_SW_Markov.cc
```diff
@@ -195,20 +191,18 @@ namespace {
...
 
-    // Reset to previous global state
-    // Reset_SOILWAT2_after_UnitTest();
     SW_MKV_deconstruct(&SW_All.Markov);
   }
 
 
   // Test correcting daily temperatures for wet/dry days
-  TEST(WGTest, WetDryTemperatureCorrection) {
+  TEST_F(AllTest, WetDryTemperatureCorrection) {
```
dschlaep, line 199: Please change `TEST_F(AllTest, WetDryTemperatureCorrection)` to something like `TEST_F(AllTest, WeatherGeneratorWetDryTemperatureCorrection)`

tests/gtests/test_SW_VegProd.cc
```diff
@@ -149,14 +146,11 @@ namespace {
...
     }
-
-    // Reset to previous global state
-    Reset_SOILWAT2_after_UnitTest();
   }
 
 
   // Test summing values across vegetation types
-  TEST(VegTest, Summing) {
+  TEST_F(AllTest, Summing) {
```
dschlaep, line 153: Please change `TEST_F(VegTest, Summing)` to something like `TEST_F(AllTest, VegProdSumming)`

tests/gtests/test_SW_VegProd.cc
```diff
@@ -174,7 +168,7 @@ namespace {
 
 
 	// Check `get_critical_rank`
-	TEST(VegTest, rank) {
+	TEST_F(AllTest, rank) {
```
dschlaep, line 171: Please change `TEST_F(VegTest, rank)` to something like `TEST_F(AllTest, VegProdCriticalRanks)`

tests/gtests/test_SW_Weather.cc
```diff
@@ -25,10 +25,10 @@
 
 namespace {
 
-    TEST(ReadAllWeatherTest, DefaultValues) {
+    TEST_F(AllTest, DefaultValues) {
```
dschlaep, line 28: Please improve test suit names here and throughout following examples from my earlier comments

dschlaep (commented):
I like having `LogInfo` as last argument and it seems the test fixtures is good progress as well!

src/SW_Site.c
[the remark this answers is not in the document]
dschlaep: I was thinking about STEPWAT2 (and maybe rSOILWAT2)

include/SW_Site.h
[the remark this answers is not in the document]
dschlaep: There are two questions
1. Why do we need these arrays to be `MAX_LAYERS + 1`?
2. If so, would something like `#define MAX_LAYERS_PLUS1 26 // MAX_LAYERS_PLUS1 = MAX_LAYERS + 1` be more efficient? Or does it not matter (much)?

src/SW_Site.c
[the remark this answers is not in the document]
N1ckP3rsl3y: Previously, `SW_SIT_clear_layers()` did not reset `SW_Site->n_layers` so I am not sure why STEPWAT2 (and maybe rSOILWAT2) would need this to happen now.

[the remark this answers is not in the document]
dschlaep: Yes, you are correct! Sorry my bad! We are and have been setting `SW_Site->n_layers = 0` in `SW_SIT_init_counts()`

include/SW_Site.h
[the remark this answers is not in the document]
N1ckP3rsl3y: Thank you for the clarification, here are my answers:

1.  This is to make room for a deep drain layer, which when running severe tests with size `MAX_LAYERS` proved that it was not sufficient space (test `SWFlowTranspWeightedAvg`, or `TranspWeightedAvg` if it's before I push my changes), hence `MAX_LAYERS + 1`.

2. The only impact on efficiency (if any) from adding something like that would be is the compilation time since this memory would be allocated then. I believe your idea would be good from the standpoint of readability (and maybe maintainability). So if we want to be descriptive, it would be a good idea to make a new define for it, like you suggested.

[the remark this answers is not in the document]
dschlaep: Ok, thanks, makes sense! Though I still not convinced that setting up the soil layer arrays for `MAX_LAYERS + 1` is the best approach -- I'll think about re-organizing deep drainage. Maybe create a dedicated attribute in "SW_SOILWAT" instead of stuffing it into `SW_SoilWat.swcBulk[][]`? Let's discuss

src/SW_Site.c
```diff
@@ -1266,52 +1288,33 @@ void nlayers_vegroots(LyrIndex n_transp_lyrs[]) {
...
+
+			SW_Site->width[SW_Site->n_layers] = 1.0;
+			SW_Site->n_layers++;
 
 			/* sp->deepdrain indicates an extra (dummy) layer for deep drainage
 			* has been added, so n_layers really should be n_layers -1
 			* NOTE: deep_lyr is base0, n_layers is BASE1
 			*/
-			SW_Site.deep_lyr = --SW_Site.n_layers;
+			SW_Site->deep_lyr = --SW_Site->n_layers;
```
dschlaep, line 1308: Why are we doing 
```
SW_Site->n_layers++;
SW_Site->deep_lyr = --SW_Site->n_layers;
```

Wouldn't the same be achieved by
```
SW_Site->deep_lyr = SW_Site->n_layers;
```

include/SW_Site.h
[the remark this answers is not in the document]
dschlaep: Resolved with commit aec25159be5d42bcd9817cc103c062010693be16 "Deep drainage is now percolation from deepest soil layer" -> no need anymore for `MAX_LAYERS + 1`

tests/gtests/sw_testhelpers.h
```diff
@@ -18,14 +23,54 @@ static const double
...
+    SW_ALL SW_All;
+    PATH_INFO PathInfo;
+    LOG_INFO LogInfo;
+    SW_OUTPUT_POINTERS SW_OutputPtrs;
+
+    void SetUp() override {
+      // `memcpy()` does not work to copy from a global to local attributes
+      // since the function does not copy dynamically allocated memory
+
+      const char * masterfile_test = "files.in"; // relative to 'dir_test'
```
dschlaep, line 39: Could we get rid of `const char * masterfile_test = "files.in";`? And directly do `PathInfo.InFiles[eFirst] = Str_Dup(DFLT_FIRSTFILE, &LogInfo);`? Would there be any downsides? This would avoid the need to maintain two instances of hard-coded "files.in".

src/SW_Flow.c
```diff
@@ -224,6 +224,11 @@ void SW_Water_Flow(SW_ALL* sw, LOG_INFO* LogInfo) {
 		);
 	}
 
+	ForEachSoilLayer(i, n_layers) {
```
dschlaep, line 227: It seems that setting the array `sw->SoilWat.avgLyrTemp[]` to 0 on each day of the simulation is unnecessary for most cases and expensive. We only need to `sw->SoilWat.avgLyrTemp[]` initialize once during a simulation run and only if the soil temperature module is turned off.

Here is my suggestion to solve this issue: https://github.com/DrylandEcology/SOILWAT2/pull/354

dschlaep:
"ubuntu-latest" GHA check is failing after commit [9216c3b](https://github.com/DrylandEcology/SOILWAT2/pull/349/commits/9216c3b43bb782d98b3730a4bfeb8b52874ad5a5) "Feature soil temperature initialization" 

See https://github.com/DrylandEcology/SOILWAT2/actions/runs/5422481107/jobs/9859172878?pr=349
with
```
make clean test_rep3rnd
...

bin/sw_test --gtest_shuffle --gtest_repeat=3
...

Repeating all tests (iteration 2) . . .

Note: Randomizing tests' orders with a seed of 19456 .

...

[ RUN      ] AllTest.ReadAllWeatherSomeMissingValuesDays
make: *** [makefile:367: test_rep3rnd] Segmentation fault (core dumped)
Error: Process completed with exit code 2.
```

Unfortunately, I can reproduce test sequence but not the error on my local machine (macOSX using clang15) with
```
CC=clang CXX=clang++ make clean_test test
bin/sw_test --gtest_shuffle --gtest_random_seed=19456
```

N1ckP3rsl3y:
I tested this on my personal Ubuntu server. After many runs (min. 50), I am unable to reproduce this problem with the same seed.

dschlaep:
The segfault on "ubuntu-latest" on GHA checks, mentioned earlier (https://github.com/DrylandEcology/SOILWAT2/pull/349#pullrequestreview-1507276530), did not reproduce even when running the same GHA check with the same seed (https://github.com/DrylandEcology/SOILWAT2/pull/355 and https://github.com/DrylandEcology/SOILWAT2/actions/runs/5426750048/jobs/9906921282?pr=355):

```
  CC=clang CXX=clang++ make clean_test test
  bin/sw_test --gtest_shuffle --gtest_random_seed=19456
...
Note: Randomizing tests' orders with a seed of 19456 .
[==========] Running 104 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 104 tests from AllTest
...
[==========] 104 tests from 1 test suite ran. (14037 ms total)
[  PASSED  ] 104 tests.
```

N1ckP3rsl3y (approved):

PR #352, opened 2023-06-23
Improve SOILWAT2 tools

dschlaep:
- script and make target to use leak sanitizer on SOILWAT2 executable running the example simulation
- GHA now include leak sanitizer for the example simulation
- improved script "check_SOILWAT2.sh": less verbose, includes leak sanitizer for the example simulation

### DIFF
--- a/tools/check_SOILWAT2.sh
+++ b/tools/check_SOILWAT2.sh
@@ -25,7 +25,7 @@ verbosity_on_error=false # options: true false
 #--- List of (builtin and macport) compilers
 
 declare -a port_compilers=(
-  "none"
+  "default compiler"
   "mp-gcc10" "mp-gcc11" "mp-gcc12"
   "mp-clang-10" "mp-clang-11" "mp-clang-12" "mp-clang-13" "mp-clang-14" "mp-clang-15" "mp-clang-16"
 )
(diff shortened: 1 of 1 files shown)


tools/check_SOILWAT2.sh
```diff
@@ -66,54 +79,145 @@ for ((k = 0; k < ncomp; k++)); do
...
+  if [ "${res_error}" ]; then
+    echo "Failure:"
+    if [ "${verbosity_on_error}" = true ]; then
+      echo "${res}"
+    else
+      echo "${res_error}"
+    fi
+
+  else
+    echo "Success: target."
```
N1ckP3rsl3y, line 149: This is nit-picky, but I have a suggestion:

Wouldn't it make it look more uniform  if successes/failures were along the lines of:
```
Result: Success
Result: Failed
```
and extend that to the other successes/failure messages? Is there a reason against this?

dschlaep: I see your point and think it works well for the targets that run the test cases. However, how would you suggest to update the output of `compare_output_against_reference()` to fit this pattern?

N1ckP3rsl3y: I see your concern with `compare_output_against_reference()`. So my suggestion is something along the lines of:

```
"Output Result: Success - Output reproduces reference exactly."
"Output Result: Failed - Output deviates beyond tolerance from reference:"
"Output Result: Success - Output reproduces reference within tolerance but not exactly:"
```
which is a bit cluttered, but does fit the pattern.

N1ckP3rsl3y (approved):

PR #355, opened 2023-06-30
Test Linux GHA

N1ckP3rsl3y:
- Ubuntu with GHA has shown that in one run, will segfault with the seed 19456
- Test GHA with the set seed of 19456


dschlaep:
Replaced by PR "Feature memory fixes" #358

PR #366, opened 2023-08-08
NetCDF Output - Create Domain netCDF

N1ckP3rsl3y:
* Second part of moving to netCDF output
* If a domain netCDF has not been provided, generate one to future runs
* See https://github.com/DrylandEcology/SOILWAT2/issues/362


include/SW_netCDF.h
```diff
@@ -0,0 +1,62 @@
...
+#define THREEDIMS 3
+
+// Constant sizes
+#define NUMBNDS 2
+
+#define SIZEONE 1
+#define SIZETWO 2
+
+void SW_NC_read_files(PATH_INFO* PathInfo, LOG_INFO* LogInfo);
+int nc_key_to_ID(char* key);
```
dschlaep, line 33: Why is the module "SW_netCDF" exporting all these functions? Most/all are only called from within the module. I think those should be static, i.e., not exported

src/SW_Carbon.c
```diff
@@ -272,8 +272,8 @@ void SW_CBN_init_run(VegType VegProd_veg[], SW_MODEL* SW_Model,
     if (SW_Carbon->use_bio_mult) {
       ForEachVegType(k) {
         VegProd_veg[k].co2_multipliers[BIO_INDEX][year] =
-                                              VegProd_veg[k].co2_bio_coeff1
-                                              * pow(ppm, VegProd_veg[k].co2_bio_coeff2);
+                            VegProd_veg[k].co2_bio_coeff1
```
dschlaep, line 275: please only do white-space adjustments when actually something in the code changes - if at all

src/SW_Domain.c
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+    #ifdef SWNETCDF
+    // Check if the provided domain netCDF does not exist and needs to be created
+    if(!FileExists(PathInfo->InFiles_nc[DOMAIN_NC])) {
+        SW_NC_create_domain(SW_Domain, PathInfo->InFiles_nc[DOMAIN_NC],
+                            LogInfo);
+    }
+    #else
+    // Default domain type to "s" and number of sites to one
```
dschlaep, line 127: Why do you overwrite `DomainType` and `nDimS` that just have been read from "domain.in" if `SWNETCDF` is not defined? This is against what we formulated in the plan.

include/SW_datastructs.h
```diff
@@ -739,7 +743,8 @@ typedef struct {
 } LOG_INFO;
 
 typedef struct {
-	char *InFiles[SW_NFILES];
+	char *InFiles_csv[SW_NFILES];
```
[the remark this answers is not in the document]
N1ckP3rsl3y: This was meant to parallel the naming of _nc variables. But I see your point, this is not meant to hold anything csv-related. I will change that back.

I was planning on splitting the variable into the two when we encounter an output file array for netCDFs so it can happen at the same time we add `OutFiles_nc`.

src/SW_Domain.c
N1ckP3rsl3y: I'm not sure I see the problem here. The program will ignore any spacial domain information when SWNETCDF is not defined. Then it will default `DomainType` to "s" and `nDimS` to 1, not overwriting but setting the variables. Then if SWNETCDF is defined, we do read in spacial domain information from `domain.in` and will not be overwritten by that chunk of code.

The plan states
> New input variables (these default to “s” and  #s = 1 if not SWNETCDF)

include/SW_datastructs.h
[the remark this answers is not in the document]
dschlaep: ok makes sense

src/SW_Domain.c
dschlaep: > New input variables (these default to “s” and #s = 1 if not SWNETCDF)

are provided as defaults via "domain.in"; they shouldn't be hard-coded

src/SW_Domain.c
```diff
@@ -0,0 +1,175 @@
...
+	f = OpenFile(MyFileName, "r", LogInfo);
+
+    // Set SW_DOMAIN
+    while(GetALine(f, inbuf)) {
+        sscanf(inbuf, "%s %s", key, value);
+
+        keyID = domain_inkey_to_id(key);
+        switch(keyID) {
+            // Only expect domain information when SWNETCDF is defined
+            #ifdef SWNETCDF
```
dschlaep, line 45: I find it confusing that `SW_DOM_read()` is not reading in domain information `#ifdef SWNETCDF` (lines 45-63) -- we just established that "domain.in" is a new text-based input for SOILWAT2 in #365.

I think we should remove those - then the else-branch on lines 126-130 becomes superfluous too (see comment https://github.com/DrylandEcology/SOILWAT2/pull/366/files#r1290876867).

src/SW_Domain.c
```diff
@@ -0,0 +1,175 @@
...
+	if (!fenddy) {
+		SW_Domain->endend = Time_get_lastdoy_y(SW_Domain->endyr);
+        LogError(LogInfo, LOGWARN,
+                "Domain.in: Missing End Day - using %d\n", SW_Domain->endend);
+	}
+
+    #ifdef SWNETCDF
+    // Check if the provided domain netCDF does not exist and needs to be created
+    if(!FileExists(PathInfo->InFiles_nc[DOMAIN_NC])) {
+        SW_NC_create_domain(SW_Domain, PathInfo->InFiles_nc[DOMAIN_NC],
```
dschlaep, line 123: We need to establish/calculate `suids` before we call `SW_NC_create_domain()` -- `suids` are one of the main pieces of information that we are writing into the "domain.nc"

src/SW_netCDF.c
```diff
@@ -0,0 +1,591 @@
...
+ *  "_ChunkSizes" attribute
+ * @param[in] chunkVals Values under the attribute "_ChunkSizes"
+ * @param[in] LogInfo Holds information dealing with logfile output
+*/
+void create_var_domain(int* domID, int domDim[], int numDims,
+        size_t chunkSize, unsigned int chunkVals[], LOG_INFO* LogInfo) {
+
+    // Longer attribute names
+    int numFillSize = 1;
+    float fillVal[] = {-3.4E38};
```
dschlaep, line 284: Let's use the "default" fill value of `NC_FILL_FLOAT` if variable type is `NC_FLOAT`. See https://docs.unidata.ucar.edu/netcdf-c/current/netcdf_8h.html#ad9a1c544d512d09c610e181f0709885a

src/SW_netCDF.c
```diff
@@ -0,0 +1,591 @@
...
+ *
+ * @param[in] nDimX Size of dimension "x"
+ * @param[in] nDimY Size of dimension "y"
+ * @param[in] domainID ID of the domain variable within the domain netCDF
+ * @param[in] xID ID of the x variable within the domain netCDF
+ * @param[in] yID ID of the y variable within the domain netCDF
+ * @param[in] x_bndsID ID of the x_bnds variable within the domain netCDF
+ * @param[in] y_bndsID ID of the y_bnds variable within the domain netCDF
+*/
+void fill_xy_vars(int nDimX, int nDimY, int domainID,
```
dschlaep, line 354: `suids`, i.e., domain, and x and y or s coordinates are calculated by "SW_domain" if needed. `fill_xy_vars()` or `fill_s_vars()` should just write coordinate values into their matching variables.

src/SW_netCDF.c
```diff
@@ -0,0 +1,591 @@
...
+
+    // Fill "domain"
+    for(indexY = 0; indexY < nDimY; indexY++) {
+        for(indexX = 0; indexX < nDimX; indexX++) {
+            setValsDom[indexY][indexX] = DomVal;
+            DomVal += DomIncrement;
+        }
+    }
+
+    nc_put_vara_float(OPEN_NC_ID, domainID, startXYDom, DomCount,
```
dschlaep, line 419: why are `suids` written as float and not as one of the (long) integer types?

src/SW_netCDF.c
```diff
@@ -0,0 +1,591 @@
...
+    sprintf(timeStr, "%d-%02d-%02d %02d:%02d:%02d",
+            timeStruct.tm_year + yearOffset, timeStruct.tm_mon + monOffset,
+            timeStruct.tm_mday, timeStruct.tm_hour, timeStruct.tm_min,
+            timeStruct.tm_sec);
+    sprintf(dateStr, "v%d%02d%02d", timeStruct.tm_year + yearOffset,
+            timeStruct.tm_mon + monOffset, timeStruct.tm_mday);
+
+    // Fill global attributes
+    write_att_str("Conventions", "CF-1.8", NC_GLOBAL, LogInfo);
+    write_att_str("created_by", "SOILWAT2 v7.1.0", NC_GLOBAL, LogInfo);
```
dschlaep, line 492: Why not use `SW2_VERSION` instead of hard coded version number?

src/SW_netCDF.c
```diff
@@ -0,0 +1,591 @@
...
+    nc_put_vara_double(OPEN_NC_ID, yID, start, count, &setValsY[0]);
+}
+
+/**
+ * @brief Fill a domain netCDF with global attributes (only when a domain
+ *  netCDF is generated)
+ *
+ * @param[in] LogInfo Holds information dealing with logfile output
+*/
+void write_global_domain_atts(LOG_INFO* LogInfo) {
```
dschlaep, line 474: let's make a generic `write_global_atts()` instead of one that is dedicated to creating "domain' only

src/SW_netCDF.c
```diff
@@ -0,0 +1,591 @@
...
+
+    nc_create(DomainName, NC_NETCDF4, NULL); // Don't store file ID (OPEN_NC_ID)
+
+    create_var_crs(LogInfo);
+    write_global_domain_atts(LogInfo);
+
+    if(strcmp(SW_Domain->DomainType, "xy") == equalString) {
+        create_dim_xy(SW_Domain->nDimX, SW_Domain->nDimY, &yDimID,
+                      &xDimID, &bndsDimID, LogInfo);
+        create_xy_vars(SW_Domain->nDimX, SW_Domain->nDimY, yDimID,
```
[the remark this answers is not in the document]
N1ckP3rsl3y: I'm not sure I understand this exactly. I am under the impression that "x" and "y" variables are the coordinates for both "xy" and "s" domain. Aside from that, I'm not sure how we would determine the coordinates. We can hard code them like what is being done in `fill_*_vars()`. Or we can get start coordinates from SW_MODEL, but that would require a dependency on SW_MODEL. And if we were to do that, when we fill all coordinate values, is the increment fixed or not? Is there something else I'm missing?

src/SW_netCDF.c
N1ckP3rsl3y: I am going off the example domain netCDFs, and since they are floats in the examples, they are being generated as floats.

src/SW_netCDF.c
```diff
@@ -0,0 +1,591 @@
...
+
+/**
+ * @brief Create an "x" variable with common attributes between the domains
+ *  "xy" and "site"
+ *
+ * @param[in] xID ID of the x variable within the domain netCDF
+ * @param[in] xDim Dimensions of the x variable
+ * @param[in] LogInfo Holds information dealing with logfile output
+*/
+void create_var_x(int* xID, int xDim[], LOG_INFO* LogInfo) {
```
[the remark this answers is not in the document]
N1ckP3rsl3y: I'm going to have to query how these values are dependent on CRS values.

src/SW_netCDF.c
```diff
@@ -0,0 +1,591 @@
...
+    int xDim[] = {xDimID}, yDim[] = {yDimID}, xBndsDim[] = {xDimID, bndsDimID},
+        yBndsDim[] = {yDimID, bndsDimID}, domainDims[] = {yDimID, xDimID};
+    int xID, yID, xBndsID, yBndsID, domID;
+    size_t numChunkVals = 2;
+    unsigned int xBndsAttVals[] = {nDimX, NUMBNDS},
+        yBndsAttVals[] = {nDimY, NUMBNDS}, ChunkVals[] = {nDimX, nDimY};
+    float fillVal[] = {-3.4E38};
+
+    // "domain" variable
+    create_var_domain(&domID, domainDims, TWODIMS, numChunkVals,
```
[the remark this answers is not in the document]
N1ckP3rsl3y: I'm not sure I understand the problem in this comment. The variable `domain` is created in these functions because when creating the variable, mainly, we can set domain-specific attributes.

I'm not sure I see what exactly, aside from `domain` itself, cannot be used outside of a domain environment (ignoring other comments). Because we have `create_variable()` and `create_dimension()` along with others which are generic. When we use `create_*_vars()` it is meant to create the entire domain file based on the domain type. With helper functions that should be generic enough to use in other occasions. Though `create_var_*()` may need changes depending on the expected attributes for output variable files.

Issue #359, opened 2023-07-28
Prevent memory leaks during error/early exist


dschlaep:
Currently, we call `LogError()` when the code encounters a problem and, if the problem is an error, then we exit early (i.e., crash for SOILWAT2 -- in case of `rSOILWAT2`, we call the `R_int.h` function `error()` instead).

However, we are not de-allocating dynamically allocated memory. For a single SOILWAT2 run, the program crashes and the OS may/should clean it up. However, in rSOILWAT2, we return to the calling function prematurely, and this causes a memory leak. Similarly, if we will convert SOILWAT2 in the future to run with multiple threads, then a prematurely returning thread will cause a memory leak.

Suggested solution: after we identify the problem but before we call `exit()` (or similar), we need to de-allocate dynamically allocated memory.

PR #367, opened 2023-08-29
Feature error handling

N1ckP3rsl3y:
* Update the logging method of errors/warnings/notes
    * Store messages throughout the program
    * No longer crash with an error
    * Make logging safe for single and muli-process/threaded environments

* This PR addresses #368 and #359


dschlaep (commented):
Great job -- lots of work!

I think we can get rid of `sw_strdup()` because the only place it is used is by `MkDir()` (plus one location in STEPWAT2) - both of which can be converted to using `Str_Dup()`

src/SW_Control.c
```diff
@@ -116,6 +116,24 @@ void SW_CTL_main(SW_ALL* sw, SW_OUTPUT_POINTERS* SW_OutputPtrs,
...
 
+/**
+ * @brief Initialize all possible pointers to NULL incase of an unexpected
+ *  program exit
+ *
+ * @param[in,out] sw Comprehensive struct of type SW_ALL containing
+ *  all information in the simulation
+ * @param[in,out] InFiles Array of program in/output files
+*/
+void SW_CTL_init_ptrs(SW_ALL* sw, char *InFiles[]) {
```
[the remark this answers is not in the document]
N1ckP3rsl3y: `SW_FILE_STATUS` - There is a chance to crash when opening the files, so set all to NULL
`SW_OUTPUT_POINTERS` - Nothing is allocated so nothing needs to be freed, no action required
`LogInfo` - A separate function will be provided in my next push

src/SW_SoilWater.c
```diff
@@ -560,6 +560,35 @@ void SW_WaterBalance_Checks(SW_ALL* sw, LOG_INFO* LogInfo)
...
+ * 		soil water related values
+*/
+void SW_SWC_init_ptrs(SW_SOILWAT* SW_SoilWat) {
+	OutPeriod pd;
+
+	// Initialize output structures:
+	ForEachOutPeriod(pd)
+	{
+		SW_SoilWat->p_accu[pd] = NULL;
+		if (pd > eSW_Day) {
```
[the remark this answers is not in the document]
N1ckP3rsl3y: We can do `SW_*_init_ptrs()` but not `SW_*_deconstruct()`. This is due to what `average_for()` does with the first period in `p_oagg`. It sets the pointer in the array of index `eSW_Day` to `p_accu`, eliminating its NULL status. Then, when we deallocate, it'll deallocate `p_accu`'s memory and try to free that instead, causing a problem. We could set `p_oagg` to NULL again (of index `eSW_Day`) once the simulation is over, but this may be unnecessary when we can keep what we currently have.

include/filefuncs.h
```diff
@@ -28,7 +28,7 @@ const char *BaseName(const char *p);
 Bool FileExists(const char *f);
 Bool DirExists(const char *d);
 Bool ChDir(const char *d);
-Bool MkDir(const char *d);
+Bool MkDir(const char *dname, LOG_INFO* LogInfo);
 Bool RemoveFiles(const char *fspec, LOG_INFO* LogInfo);
 void sw_error(int errorcode, const char *format, ...);
```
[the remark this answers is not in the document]
N1ckP3rsl3y: We can make it a private, non-static function. The compiler flag `[-Wunused-function]` has an issue when it is static and not used (since we will only use it in STEPWAT2).

tests/gtests/sw_maintest.cc
```diff
@@ -1,4 +1,4 @@
-#include "gtest/gtest.h"
+#include <gmock/gmock.h>
```
dschlaep, line 1: We'll need to update `makefile` so that compilation can find "include/googletest/googlemock/"

include/filefuncs.h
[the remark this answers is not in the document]
dschlaep: See my other comment about removing `sw_error()` -- if not then we should wrap `static void sw_error()` in `#if defined (STEPWAT)` directives - exactly because it is unused in SOILWAT code

src/SW_Main_lib.c
```diff
@@ -191,19 +191,98 @@ void sw_init_args(int argc, char **argv, Bool *QuietMode,
...
+}
+
+/**
+ * @brief Write logs that have been accumulated throughout the program/
+ * 	simulation run
+ *
+ * @param[in] QuietMode Flag to control if the program version is displayed
+ * @param[in] LogInfo Holds information dealing with logfile output
+*/
+void sw_write_logs(Bool QuietMode, LOG_INFO* LogInfo) {
```
dschlaep, line 238: In `sw_write_logs()` if `RSOILWAT`, we currently call `warning()` if we got a warning message - previously, we had functionality (in `LogError()`) that didn't create warning messages if the user requested silence (see old code comment "// send warning message only if not silenced (fp is not NULL)". Please add corresponding functionality back in.

src/SW_Model.c
```diff
@@ -113,7 +113,7 @@ void SW_MDL_read(SW_MODEL* SW_Model, char *InFiles[], LOG_INFO* LogInfo) {
 	FILE *f;
 	int y, cnt;
 	TimeInt d;
-	char *p, enddyval[6], errstr[MAX_ERROR], *MyFileName, inbuf[MAX_FILENAMESIZE];
+	char *p, enddyval[6], errstr[MAX_LOG_SIZE], *MyFileName, inbuf[MAX_FILENAMESIZE];
```
dschlaep, line 116: see earlier comment about `MAX_LOG_SIZE` - I think we should create local `errstr[]` of `SW_MDL_read()` to be shorter than `MAX_LOG_SIZE`

src/filefuncs.c
```diff
@@ -146,67 +108,44 @@ void LogError(LOG_INFO* LogInfo, const int mode, const char *fmt, ...) {
...
-        strcpy(outfmt, "");
-    else if (LOGNOTE & mode)
-        strcpy(outfmt, "NOTE: ");
-    else if (LOGWARN & mode)
+    if (LOGWARN & mode)
         strcpy(outfmt, "WARNING: ");
     else if (LOGERROR & mode)
         strcpy(outfmt, "ERROR: ");
 
     strcat(outfmt, fmt);
```
dschlaep, line 123: In `LogError()`, we are currently `strcat()` the "program" provided message into our local buffer `outfmt` -- to me this looks that we could potentially write out of buffer (if "program" message happens to be longer than `outfmt`):
```
    strcat(outfmt, fmt);
    strcat(outfmt, "\n");
```

I think we should use something safer, e.g., `snprintf(outfmt, sizeof outfmt, "%s\n", ftm)` - what do you think?

N1ckP3rsl3y: I can understand where the use of `snprintf()` in this case comes from. But I feel this is not necessary because
* After checking the format messages in SOILWAT2, the newest value of `MAX_LOG_SIZE` (400) may in itself be considered overkill. That is, there shouldn't be a reason to exceed the size of `outfmt`. And briefly looking at STEPWAT2, it should be a similar situation, staying below the maximum characters.
* I feel the message should be displayed in full if it were to reach the maximum buffer capacity. So, instead of cutting off the message, other things should change before taking that action. This is, of course, assuming cutting off the message is not a desired feature.

dschlaep: What happens with `strcat(outfmt, fmt);` when a future programmer would add a long message, e.g.,
```
LogError(LogInfo, LOGWARN, 
    "This message is going to be longer than 400 characters: "
    "massa tincidunt nunc pulvinar sapien et ligula ullamcorper malesuada proin libero nunc consequat interdum varius sit amet mattis vulputate enim nulla aliquet porttitor lacus luctus accumsan tortor posuere ac ut consequat semper viverra nam libero justo laoreet sit amet cursus sit amet dictum sit amet justo donec enim diam vulputate ut pharetra sit amet aliquam id diam maecenas ultricies mi eget mauris pharetra et ultrices neque ornare aenean euismod elementum nisi quis eleifend quam adipiscing vitae proin sagittis nisl rhoncus mattis rhoncus urna neque viverra justo nec ultrices dui sapien eget mi proin sed libero enim sed faucibus turpis in"
);
```

or inject values that end up being many characters long, e.g., 
```
const char *buf = "massa tincidunt nunc pulvinar sapien et ligula ullamcorper malesuada proin libero nunc consequat interdum varius sit amet mattis vulputate enim nulla aliquet porttitor lacus luctus accumsan tortor posuere ac ut consequat semper viverra nam libero justo laoreet sit amet cursus sit amet dictum sit amet justo donec enim diam vulputate ut pharetra sit amet aliquam id diam maecenas ultricies mi eget mauris pharetra et ultrices neque ornare aenean euismod elementum nisi quis eleifend quam adipiscing vitae proin sagittis nisl rhoncus mattis rhoncus urna neque viverra justo nec ultrices dui sapien eget mi proin sed libero enim sed faucibus turpis in";

LogError(LogInfo, LOGWARN, "This message is going to be longer than 400 characters: %s\n.", buf);
```

Wouldn't this result in a buffer overflow?

include/SW_Defines.h
```diff
@@ -100,6 +100,9 @@ extern "C" {
 /* Maximum number of messages to store in LOG_INFO */
 #define MAX_MSGS 10
 
+/* Maximum number of characters for a single message in LOG_INFO */
+#define MAX_LOG_SIZE 400
```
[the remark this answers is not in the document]
dschlaep: To me this looks like unsafe code previously with `MAX_ERROR` and now with `MAX_LOG_SIZE`.

src/SW_Main_lib.c
```diff
@@ -240,26 +244,32 @@ void sw_write_logs(Bool QuietMode, LOG_INFO* LogInfo) {
...
+                          " only %d were printed.\n";
 
 	if(warningUpperBound > MAX_MSGS) {
 		warningUpperBound = MAX_MSGS;
 		tooManyWarns = swTRUE;
 
-        sprintf(tooManyWarnsStr, "There were a total of %d warnings and"\
-                                 " only %d were printed.\n",
-                                 LogInfo->numWarnings, MAX_MSGS);
+        snprintf(tooManyWarnsStr, sizeof(tooManyStrFmt), tooManyStrFmt,
```
dschlaep, line 254: `snprintf(tooManyWarnsStr, sizeof(tooManyStrFmt), ...` doesn't do the correct thing. This needs to be `snprintf(tooManyWarnsStr, sizeof(tooManyWarnsStr), ...` instead.

For instance, 
```
snprintf(tooManyWarnsStr, sizeof(tooManyStrFmt), tooManyStrFmt, 1000000, MAX_MSGS);
```
will cut off several of the last characters of `tooManyWarnsStr` - this is because the values get injected after `sizeof(tooManyStrFmt)` is determined.

include/SW_Defines.h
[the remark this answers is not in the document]
dschlaep: addressed via commit 9e8daee725d35f47df5900a30c15d8a93ab22e87 "Safe copy message format in `LogError()`"

src/SW_Main_lib.c
```diff
@@ -191,18 +195,110 @@ void sw_init_args(int argc, char **argv, Bool *QuietMode,
...
+/**
+ * @brief Initialize values within LOG_INFO
+ *
+ * @param[in] logInitPtr Log file pointer that LOG_INFO will be initialized to
+ * @param[in,out] LogInfo Holds information dealing with log file output
+*/
+void sw_init_logs(FILE* logInitPtr, LOG_INFO* LogInfo) {
+
+	LogInfo->logged = swFALSE;
+	LogInfo->logfp = logInitPtr;
```
dschlaep, line 231: Currently, we are not deleting/removing the logfile content from a previous SOILWAT2 run. For instance, we run SOILWAT2 and this leads to an error and/or warning, then that error and/or warning text is written to the logfile. We fix the inputs and now SOILWAT2 runs without errors and warnings, but the logfile still shows the error and/or warning from the previous run.

We need to reset/delete the content of the logfile, e.g., in `SW_F_read()` -- although, I thought that ``LogInfo->logfp = OpenFile(PathInfo->InFiles[eLog], "w", LogInfo);` should reset the logfile content based on https://en.cppreference.com/w/c/io/fopen : file access mode "w", then action if file already exists is "destroy content". Why doesn't it?

dschlaep: Ah, my text editor didn't update cached content if content was deleted... all good

src/filefuncs.c
```diff
@@ -75,11 +94,27 @@ static char **getfiles(const char *fspec, int *nfound, LOG_INFO* LogInfo) {
 			(*nfound)++;
 			if (alloc) {
 				flist = (char **) Mem_ReAlloc(flist, sizeof(char *) * (*nfound), LogInfo);
+                flist[(*nfound) - 1] = NULL;
```
dschlaep, line 97: Re commit cafc45b5156470637ee15e654482d4bd10d6c4fe, in `getfiles()`:

`flist[(*nfound) - 1] = NULL;` doesn't work if the precedent call to `flist = Mem_ReAlloc(...)` or `flist = Mem_Malloc(...)` fails. If it fails, then `flist` is NULL and cannot be subset (undefined behavior) and `freeGetfilesEarly()` does not free memory that was previously allocated to `flist` before it was overwritten to NULL.

PR #370, opened 2023-09-11
Feature Individual Error Handling

N1ckP3rsl3y:
- Extends https://github.com/DrylandEcology/SOILWAT2/pull/367
- Properly handles a thrown error
    - Provides premature function exiting when an error is encountered
    - Guarantee memory will be freed and files are closed upon error
- Addresses https://github.com/DrylandEcology/SOILWAT2/issues/359

### DIFF
--- a/src/SW_Markov.c
+++ b/src/SW_Markov.c
@@ -160,6 +160,7 @@ static void mvnorm(RealD *tmax, RealD *tmin, RealD wTmax, RealD wTmin,
 
 	if (GT(s, wTmin_var)) {
 		LogError(LogInfo, LOGERROR, "\nBad covariance matrix in mvnorm()");
+        return; // Exit function prematurely due to error
 	}
 
 	/* Apparently, it's possible for some but not all setups that
@@ -233,18 +234,39 @@ void SW_MKV_construct(unsigned long rng_seed, SW_MARKOV* SW_Markov,
 
 	SW_Markov->wetprob = (RealD *)
 						Mem_Calloc(MAX_DAYS, s, "SW_MKV_construct", LogInfo);
+    if(LogInfo->stopRun) {
+        return; // Exit function prematurely due to error
+    }
 	SW_Markov->dryprob = (RealD *)
 						Mem_Calloc(MAX_DAYS, s, "SW_MKV_construct", LogInfo);
+    if(LogInfo->stopRun) {
+        return; // Exit function prematurely due to error
+    }
 	SW_Markov->avg_ppt = (RealD *)
 						Mem_Calloc(MAX_DAYS, s, "SW_MKV_construct", LogInfo);
+    if(LogInfo->stopRun) {
+        return; // Exit function prematurely due to error
+    }
 	SW_Markov->std_ppt = (RealD *)
 						Mem_Calloc(MAX_DAYS, s, "SW_MKV_construct", LogInfo);
+    if(LogInfo->stopRun) {
+        return; // Exit function prematurely due to error
+    }
 	SW_Markov->cfxw = (RealD *)
 						Mem_Calloc(MAX_DAYS, s, "SW_MKV_construct", LogInfo);
+    if(LogInfo->stopRun) {
+        return; // Exit function prematurely due to error
+    }
 	SW_Markov->cfxd = (RealD *)
 						Mem_Calloc(MAX_DAYS, s, "SW_MKV_construct", LogInfo);
+    if(LogInfo->stopRun) {
+        return; // Exit function prematurely due to error
+    }
 	SW_Markov->cfnw = (RealD *)
 						Mem_Calloc(MAX_DAYS, s, "SW_MKV_construct", LogInfo);
+    if(LogInfo->stopRun) {
+        return; // Exit function prematurely due to error
+    }
 	SW_Markov->cfnd = (RealD *)
 						Mem_Calloc(MAX_DAYS, s, "SW_MKV_construct", LogInfo);
 }
@@ -427,6 +449,7 @@ Bool SW_MKV_read_prob(char *InFiles[], SW_MARKOV* SW_Markov,
 					" line %d of file %s\n",
 					lineno,
 					MyFileName);
+            return swFALSE; // Exit function prematurely due to error
 		}
 
 		// Check that input values meet requirements:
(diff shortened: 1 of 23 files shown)


src/SW_Output.c
```diff
@@ -274,6 +274,8 @@ static void sumof_vpd(SW_VEGPROD *v, SW_VEGPROD_OUTPUTS *s, OutKey k, TimeInt do
 
 		default:
 			LogError(LogInfo, LOGERROR, "PGMR: Invalid key in sumof_vpd(%s)", key2str[k]);
+            return; // Exit function prematurely due to error
```
dschlaep, line 277: Why did the functions `str2key()` and `str2stype()` not obtain a `return; // Exit function prematurely due to error`? I agree it is not strictly -- I was just guessing that you added a premature `return` to every instance of `LogError(LogInfo, LOGERROR, ...)` for consistency?

src/SW_Site.c
```diff
@@ -430,6 +430,7 @@ void SWRC_PTF_estimate_parameters(
 			LOGERROR,
 			"PTF is not implemented in SOILWAT2."
 		);
+        return; // Exit function prematurely due to error
 	}
 
 	/**********************************/
```
dschlaep, line 436: Why did the function `SW_swcBulk_saturated()` not get a premature `return` in 
```
default:
			LogError(
				LogInfo,
				LOGERROR,
				"`SW_swcBulk_saturated()`: SWRC (type %d) is not implemented.",
				swrc_type
			);
			break;
```
?

src/SW_Site.c
```diff
@@ -630,6 +631,7 @@ double SW_swcBulk_minimum(
 				"`SW_swcBulk_minimum()`: SWRC (type %d) is not implemented.",
 				swrc_type
 			);
+            return theta_min_theoretical; // Exit function prematurely due to error
```
dschlaep, line 634: Let's not return values that could be confused with real values on error
```
LogError(
				LogInfo,
				LOGERROR,
				"`SW_swcBulk_minimum()`: SWRC (type %d) is not implemented.",
				swrc_type
			);
            return theta_min_theoretical; // Exit function prematurely due to error
```
Let's return `SW_MISSING` (as done elsewhere)

src/SW_Weather.c
```diff
@@ -1746,6 +1774,9 @@ void SW_WTH_read(SW_WEATHER* SW_Weather, SW_SKY* SW_Sky, SW_MODEL* SW_Model,
```
dschlaep, line 1774: Why is `check_and_update_dailyInputFlags()` not returning prematurely on error?

tests/gtests/sw_testhelpers.cc
```diff
@@ -43,6 +43,10 @@ void create_test_soillayers(unsigned int nlayers,
       "requested number of soil layers (n = %d) is not accepted.\n", nlayers);
   }
 
+  if(LogInfo->stopRun) {
+    sw_check_exit(swFALSE, LogInfo);
```
dschlaep, line 47: In `create_test_soillayers()` for unit tests, why `sw_check_exit(swFALSE, LogInfo);` instead of recoiling early as elsewhere?

src/SW_Markov.c
dschlaep, line 1: why doesn't `SW_MKV_today()` have a early return on error after the call to `mvnorm(..., LogInfo)`?

src/SW_Output_outtext.c
dschlaep, line 1: Let's add early returns in STEPWAT functions, e.g., `setGlobalSTEPWAT2_OutputVariables()`, `_create_filename_ST()`, `_create_csv_file_ST()`, etc. -- it doesn't harm (because we currently crash in LogError()` and it avoids laboriously fixing SOILWAT2 repository code, just in case STEPWAT switches in the future to early recoil

src/SW_Site.c
```diff
@@ -2052,6 +2086,7 @@ void SW_SIT_init_run(SW_VEGPROD* SW_VegProd, SW_SITE* SW_Site, LOG_INFO* LogInfo
 				"Soil density type not recognized",
```
dschlaep, line 2086: why doesn't `SW_SIT_init_run()` return early on error after calls `calculate_soilMatricDensity(..., LogInfo)`, `SWRC_PTF_estimate_parameters(..., LogInfo)`, `SW_SWRC_SWPtoSWC(..., LogInfo)`, `SW_swcBulk_saturated()`, `SW_swcBulk_minimum()` etc

src/SW_Weather.c
dschlaep, line 1: `readAllWeather()` should return early on error after `_read_weather_hist()` -- since this is called in a loop

src/SW_Weather.c
```diff
@@ -882,6 +885,7 @@ void generateMissingWeather(
       "generateMissingWeather(): method = %u is not implemented.\n",
```
dschlaep, line 885: `generateMissingWeather()` should return early on error after `SW_MKV_today()` -- since this is called in a loop

src/SW_SoilWater.c
```diff
@@ -304,7 +304,7 @@ static double itp_FXW_for_phi(double theta, double *swrcp, LOG_INFO* LogInfo) {
 
 
 	#ifdef SWDEBUG
-	if (debug) {
+	if (debug && !LogInfo->stopRun) {
```
dschlaep, line 307: please remove the ` && !LogInfo->stopRun` from `if (debug && !LogInfo->stopRun) {` in "SWDEBUG" section of `itp_FXW_for_phi()` -- that debug output was intended to be printed on error (see the else branch of `!missing(phi)`)

src/filefuncs.c
```diff
@@ -46,6 +46,9 @@ static char **getfiles(const char *fspec, int *nfound, LOG_INFO* LogInfo) {
 
 	DirName(fspec, dname); // Copy `fspec` into `dname`
 	fname = Str_Dup(BaseName(fspec), LogInfo);
+    if(LogInfo->stopRun) {
```
dschlaep, line 49: In `getfiles()`, we need to move `(*nfound) = 0;` (now line 63) to before the first early return on error (now line 50). Otherwise, `RemoveFiles()` will get whatever value for `nfiles` when it enters `for (i = 0; i < nfiles; i++) Mem_Free(flist[i]);` and possibly fails.

dschlaep (approved):

Issue #360, opened 2023-08-03
New "domain" that locates a run in a larger set of runs


dschlaep:
- Capability to define a simulation domain, i.e., a set that describes one or many simulation runs
  - A domain may be composed of gridcells or a list of sites (all domain related sizes need to be long int because we want to possible represent more than 2.1e9 elements)
  - The spatial and temporal information of the domain defines the spatial and temporal structure of (future) netCDF inputs and outputs
  - Avoid storing full domain information in memory (large simulation sets will be too large to hold in memory)

- Loop over simulation domain/set
  - Loop over all (uncompleted) simulation runs in domain
  - Alternatively, simulate one user-specified gridcell/site in the domain (at a time)

- Do not crash/exit during individual simulation runs but collect, clean up, and report for the entire set/domain

- Full backwards compatibility

PR #377, opened 2023-10-20
Domain - set of simulation runs

N1ckP3rsl3y:
* Prepare SOILWAT2 to interact with netCDFs in the future
* Handle domain-related changes'
* See https://github.com/DrylandEcology/SOILWAT2/issues/360
* This PR is a continuance of https://github.com/DrylandEcology/SOILWAT2/pull/365 due to branch rename


src/SW_Domain.c
```diff
@@ -0,0 +1,265 @@
...
+        if(userSUID >= nSUIDs) {
+            LogError(LogInfo, LOGERROR, "User input for start suid is bigger"
+                              " than the total number of suids.");
+            return; // Exit function prematurely due to error
+        }
+
+        *startSimSet = userSUID;
+        *endSimSet = userSUID + 1;
+    } else {
+        *endSimSet = nSUIDs;
```
dschlaep, line 225: `make test_severe` suggests an error with `*endSimSet = nSUIDs;`

```
src/SW_Domain.c:225:9: runtime error: store to null pointer of type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/SW_Domain.c:225:9 in 
AddressSanitizer:DEADLYSIGNAL
=================================================================
==73763==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x00010343d6c0 bp 0x00016cf1eb00 sp 0x00016cf1e930 T0)
==73763==The signal is caused by a UNKNOWN memory access.
==73763==Hint: address points to the zero page.
    #0 0x10343d6c0 in SW_DOM_SimSet SW_Domain.c:225
...
```

src/SW_Main.c
```diff
@@ -68,63 +75,93 @@ int main(int argc, char **argv) {
...
+        ncStartSuid = SW_DOM_calc_ncStartSuid(&SW_Domain, suid);
+
+        if(SW_DOM_CheckProgress(SW_Domain.DomainType, ncStartSuid)) {
+            SW_CTL_run_sw(&sw_template, &SW_Domain, ncStartSuid, NULL,
+                          SW_OutputPtrs, NULL, &LogInfo);
+
+            sw_write_warnings(&LogInfo);
+            sw_init_logs(stdout, &LogInfo); // Reset LOG_INFO for next simulation run
+
+            if(!LogInfo.stopRun) {
```
dschlaep, line 146: We need early returns on error in the domain loop

src/SW_Main.c
```diff
@@ -68,63 +75,93 @@ int main(int argc, char **argv) {
...
 
-  // run simulation: loop through each year
-	SW_CTL_main(&sw, SW_OutputPtrs, &LogInfo);
+  // run simulations: loop over simulation set
+    for(suid = startSimSet; suid < endSimSet; suid++)
+    {
+        ncStartSuid = SW_DOM_calc_ncStartSuid(&SW_Domain, suid);
+
+        if(SW_DOM_CheckProgress(SW_Domain.DomainType, ncStartSuid)) {
+            SW_CTL_run_sw(&sw_template, &SW_Domain, ncStartSuid, NULL,
```
dschlaep, line 140: The domain loop currently fails silently if the domain is larger than one

src/SW_Main_lib.c
```diff
@@ -194,6 +197,17 @@ void sw_init_args(int argc, char **argv,
                     }
 					break;
 
+                case 6: /* -s */
+                    *userSUID = atoi(str) - 1; /* Make suid base0 */
+
+                    if(*userSUID < 0) {
```
dschlaep, line 203: Any value for "s" triggers the error "Input suid was less than 1. Please input a value >= 1."
e.g.,
`bin/SOILWAT2 -d tests/example/ -f files.in -s 1`

dschlaep (commented):
Great progress, thanks! Some comments. It doesn't quite work yet, let's think a little more about the consequences of implementing a simulation domain.

include/SW_Domain.h
```diff
@@ -0,0 +1,26 @@
...
+/* =================================================== */
+/*                  Local Definitions                  */
+/* --------------------------------------------------- */
+int* SW_DOM_calc_ncStartSuid(SW_DOMAIN* SW_Domain, int suid);
+void SW_DOM_calc_nSUIDs(SW_DOMAIN* SW_Domain);
+Bool SW_DOM_CheckProgress(char* domainType, int* ncStartSuid);
+void SW_DOM_CreateProgress(SW_DOMAIN* SW_Domain);
+void SW_DOM_read(char *InFiles[], SW_DOMAIN* SW_Domain, LOG_INFO* LogInfo);
+void SW_DOM_SetProgress(char* domainType, int* ncStartSuid);
+void SW_DOM_SimSet(SW_DOMAIN* SW_Domain, int userSUID, int nSUIDs,
```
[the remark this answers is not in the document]
N1ckP3rsl3y: I believe we need to have a quick discussion on what to expect from the user suid. Because if this is an unsigned long (which is in my latest push), we cannot really detect a value less than zero unless we read in a long, check then convert it to unsigned. However, this will limit user input values compared to what the program uses.

Issue #379, opened 2023-10-26
Spinup before simulation loop


dschlaep:
Implement spin-up capabilities

- before the simulation loops over years and days
- run a “spin-up” (without producing “outputs”) and pass resulting states (values of soil moisture, soil temperature, etc.) into the simulation loop
- Inputs to the “spin-up” consist of a subset of inputs for the simulation; the subset is determined by new user inputs
  - spin-up scope: use first n year(s) of simulation inputs
  - spin-up duration: sample m year(s) out of the first n year(s)
  - spin-up mode
    - randomly sample m out of the first n years
      - “repeated random resample”
  - use a sequence of m years out of the first n years, i.e., 1:m if m <= n else {1:n, 1:n, …}_m
    - “repeated sequence subset”
  - rng seed (0 means use current time, i.e., no replicable)
- defaults (backwards compatible, i.e., no spin-up): scope n = 1, duration m = 0, mode = “repeated random resample”, seed = 11

PR #385, opened 2023-11-30
Feature SpinUp

niteflyunicorns:
SpinUp feature functionality. 
Addreses #379.

### DIFF
--- a/src/SW_Control.c
+++ b/src/SW_Control.c
@@ -537,7 +537,7 @@ void SW_CTL_run_current_year(SW_ALL* sw, SW_OUTPUT_POINTERS* SW_OutputPtrs,
 */
 void SW_CTL_run_spinup(SW_ALL* sw, LOG_INFO* LogInfo) {
   
-  unsigned int i, quotient = 0, remainder = 0;
+  unsigned int i, k, quotient = 0, remainder = 0;
   int mode = sw->Model.spinup_mode,
       yr;
   TimeInt
@@ -574,7 +574,8 @@ void SW_CTL_run_spinup(SW_ALL* sw, LOG_INFO* LogInfo) {
           years[ i ] = yr + ( i % scope );
         }
         for ( i = 0; i < remainder; i++) {
-          years[ i + ( scope * quotient ) ] = yr + i;
+          k = i + (scope * quotient);
+          years[ k ] = yr + i;
         }
       }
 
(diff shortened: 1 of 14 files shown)


src/SW_Domain.c
```diff
@@ -156,6 +156,57 @@ void SW_DOM_read(SW_DOMAIN* SW_Domain, LOG_INFO* LogInfo) {
                                 Time_get_lastdoy_y(SW_Domain->endyr) : 365;
                 fenddy = swTRUE;
                 break;
+            case 8: // Spinup Mode
+                y = atoi(value);
+                
+                if (y != 1 || y != 2) {
```
dschlaep, line 162: The check against bad user input for spinup mode `y != 1 || y != 2` is always true -- see "warning: overlapping comparisons always evaluate to true [-Wtautological-overlap-compare]"
For example, if y = 1, then y != 2 is true; if y = 2, then y != 1 is true; if y = 3, then both y != 1 and y != 2 are true.

It appears that the desired logic should be `!(y == 1 || y == 2)` or (equivalently) `y != 1 && y != 2`

src/SW_Control.c
```diff
@@ -518,6 +525,86 @@ void SW_CTL_run_current_year(SW_ALL* sw, SW_OUTPUT_POINTERS* SW_OutputPtrs,
...
+          years[ i + ( scope * quotient ) ] = yr + i;
+        }
+      }
+
+      break;
+    default: // same as case 1
+      // initialize random array
+      for ( i = 0; i < duration; i++ ) {
+        yr = RandUniIntRange( sw->Model.startyr,
+                                      finalyr,
```
dschlaep, line 584: We define `spinup_scope` as "first N years of simulation" ; however, we set `finalyr = sw->Model.startyr + scope` and `RandUniIntRange()` includes the bounds. This results that mode 1 uses the first N + 1 years. Thus, we need `yr = RandUniIntRange( sw->Model.startyr, finalyr - 1, ...)` -- or `finalyr = sw->Model.startyr + scope - 1`.

src/SW_Control.c
```diff
@@ -518,6 +525,92 @@ void SW_CTL_run_current_year(SW_ALL* sw, SW_OUTPUT_POINTERS* SW_OutputPtrs,
...
+          as specified by the given spinup scope and duration which
+          then calls 'SW_SWC_water_flow' for each day.
+
+@param[in,out] sw Comprehensive struct of type SW_ALL containing all
+  information in the simulation
+@param[in,out] SW_OutputPtrs SW_OUTPUT_POINTERS of size SW_OUTNKEYS which
+  hold pointers to subroutines for output keys
+@param[in,out] LogInfo Holds information dealing with logfile output
+*/
+void SW_CTL_run_spinup(SW_ALL* sw, SW_OUTPUT_POINTERS* SW_OutputPtrs,
```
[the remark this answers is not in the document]
niteflyunicorns: I agree this is a good course of action. I am including it in my next round of edits.

dschlaep (commented):
Looks great, thanks for addressing my comments!

Issue #387, opened 2024-01-04
Track progress when simulating a domain


dschlaep:
- create a progress netCDF
- report progress for each suid
- simulate only incomplete suids (this allows re-starts)

PR #388, opened 2024-01-05
Create a "Progress" netCDF

N1ckP3rsl3y:
* Third part in the sequence of moving to netCDFs
* Create a netCDF file - "progress.nc" to keep track of the simulations that have run, won't be run, and are set to run
* See issue #387

### DIFF
--- a/src/SW_netCDF.c
+++ b/src/SW_netCDF.c
@@ -1947,18 +1947,25 @@ void SW_NC_deconstruct(SW_NETCDF* SW_netCDF) {
  *  netCDF file information
 */
 void SW_NC_open_files(SW_NETCDF* SW_netCDF, LOG_INFO* LogInfo) {
-    int fileNum, openType;
+    int fileNum, openType, *fileID;
+    char* fileName;
 
     for(fileNum = 0; fileNum < SW_NVARNC; fileNum++) {
-        if(FileExists(SW_netCDF->InFilesNC[fileNum])) {
+        fileName = SW_netCDF->InFilesNC[fileNum];
+        fileID = &SW_netCDF->ncFileIDs[fileNum];
+
+        if(FileExists(fileName)) {
             openType = (fileNum == vNCprog) ? NC_WRITE : NC_NOWRITE;
 
-            if(nc_open(SW_netCDF->InFilesNC[fileNum], openType,
-                                    &SW_netCDF->ncFileIDs[fileNum]) != NC_NOERR) {
-
+            if(nc_open(fileName, openType, fileID) != NC_NOERR) {
                 LogError(LogInfo, LOGERROR, "An error occurred when opening %s.",
-                                            SW_netCDF->InFilesNC[fileNum]);
+                                            fileName);
+                return; // Exit function prematurely due to error
+            }
 
+            get_var_identifier(*fileID, SW_netCDF->varNC[fileNum],
+                               &SW_netCDF->ncVarIDs[fileNum], LogInfo);
+            if(LogInfo->stopRun) {
                 return; // Exit function prematurely due to error
             }
         }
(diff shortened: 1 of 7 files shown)


src/SW_netCDF.c
```diff
@@ -1896,16 +2265,25 @@ void SW_NC_deconstruct(SW_NETCDF* SW_netCDF) {
...
-        if(FileExists(SW_netCDF->InFilesNC[fileNum])) {
-            if(nc_open(SW_netCDF->InFilesNC[fileNum], NC_NOWRITE,
-                                    &SW_netCDF->ncFileIDs[fileNum]) != NC_NOERR) {
+        fileName = SW_netCDF->InFilesNC[fileNum];
+        fileID = &SW_netCDF->ncFileIDs[fileNum];
+
+        if(FileExists(fileName)) {
+            openType = (fileNum == vNCprog) ? NC_WRITE : NC_NOWRITE;
 
+            if(nc_open(fileName, openType, fileID) != NC_NOERR) {
```
dschlaep, line 2278: How cumbersome would it be to enable a user that wants the "progress" variable to be added to the "domain.nc"?

Currently, if I change "files_nc.in" to `progress progress Input_nc/domain.nc`, then `nc_open(fileName, openType, fileID)` fails with error code NC_EHDFERR (-101) for "progress".

dschlaep (commented):
Works great, thanks! Just one question

src/SW_netCDF.c
N1ckP3rsl3y: This is a good point. The error is most likely coming from the fact that `domain.nc` would be opened in more than one mode `NC_NOWRITE` and `NC_WRITE`. So to answer the question, it would depend on how it is implemented.

* Most likely, the best way would be to just open all input files with purpose to write to it. Then we modify the testing of the template/progress variable and add the variable if needed. And since this most likely opens up two file handlers, at the end we could close the progress ID and swap it with the domain ID if they are the same.

There are most likely other ways of handling this, but I'd say we discuss those if needed.

dschlaep: This works nicely, thanks!

src/SW_netCDF.c
```diff
@@ -2024,6 +2030,11 @@ void SW_NC_create_progress(SW_DOMAIN* SW_Domain, LOG_INFO* LogInfo) {
             create_full_var(progFileID, NC_BYTE, 0, 0, progVarName,
                             attNames, attVals, numAtts, LogInfo);
         } else {
+            #if defined(SOILWAT)
+            if(LogInfo->printProgressMsg) {
+                sw_message("is creating a progress tracker ...");
```
N1ckP3rsl3y, line 2035: Shouldn't this message (or two similar ones) happen when we create a progress netCDF file and when we add a progress variable to the domain netCDF file?

dschlaep: Ah, yes, thanks! -- The location for the progress tracker messages was from before commit 42cc205f02bc4945e061de93de4e6dae01aeed76 "Allow for progress variable to be within domain netCDF". I'll fix this.
`SW_NC_create_domain_template()` contains a progress message for the domain netCDF template -- are you thinking about the need for an additional message?

N1ckP3rsl3y: Makes sense, thanks! The only other message I was possibly thinking of, would be something along the lines of `"is adding a progress tracker in the domain file..."` when we need to add the variable to the domain netCDF. And modifying the current on to something like `"is creating a progress tracker in it's own file..."`, to be a bit more descriptive. What do you think?

dschlaep: Yes, good idea, though I prefer to keep the use of `sw_message()` as simple as possible and console output as lean as reasonable. Ok with you?

N1ckP3rsl3y: I have no problems with this, so it sounds good to me.

Issue #412, opened 2024-06-18
Rename `SW_ALL` to `SW_RUN`


N1ckP3rsl3y:
* Rename `SW_ALL` to `SW_RUN`

Reason: The structure `SW_DOMAIN` is meant to contain information that lives above all simulation runs. While `SW_ALL` is passed around/used within the simulation runs, so `SW_RUN` is a more proper name.

Issue #395, opened 2024-04-29
Separate domain-level vs. run-specific output information


dschlaep:
Restructure code to separate domain-level vs. run-specific output information across structures `SW_OUTPUT`, `SW_GEN_OUT`, and `SW_OUTPUT_POINTERS`.

For instance, `SW_GEN_OUT` contains
    * `tOffset` which changes with each run (during the run)
    * `colnames_OUT` which are identical for all simulation runs

--> Run-specific information needs to live somewhere inside `SW_ALL`
--> Domain-level information needs to live somewhere inside `SW_DOMAIN`

PR #415, opened 2024-07-03
Separate domain- and simulation-level output information

N1ckP3rsl3y:
Reduce the copying of domain-specific information throughout simulation runs achieved by the changes

* Rename SW_ALL to SW_RUN to clarify that the information within it is simulation-dependent (see #412)
* Remove SW_OUTPUT, SW_OUTPUT_POINTERS, SW_GEN_OUT
* Create two new structs: SW_OUT_DOM and SW_OUT_RUN (see #395)


src/SW_Output.c
```diff
@@ -1091,36 +1050,36 @@ static void collect_sums(
 
 #ifdef STEPWAT
 static void _set_SXWrequests_helper(
+    SW_OUT_DOM *OutDom,
```
N1ckP3rsl3y, line 1053: The interface for `_set_SXWrequests_helper()` and `SW_OUT_set_SXWrequests()` can be simplified to no longer explicitly mention `timeSteps_SXW`.

tests/gtests/sw_testhelpers.cc
```diff
@@ -13,9 +13,8 @@
 #include <string.h>             // for strcpy
 
 
-SW_ALL template_SW_All;
+SW_RUN template_SW_All;
```
dschlaep, line 16: Let's rename `template_SW_All` to `template_SW_Run`

tests/gtests/sw_testhelpers.h
```diff
@@ -55,23 +54,36 @@ void teardown_testGlobalSoilwatTemplate();
 */
 class AllTestFixture : public ::testing::Test {
   protected:
-    SW_ALL SW_All;
+    SW_RUN SW_All;
```
dschlaep, line 57: Let's rename `SW_All` to `SW_Run`

src/SW_Output_get_functions.c
```diff
@@ -169,34 +197,45 @@ void get_co2effects_text(OutPeriod pd, SW_ALL *sw) {
...
+    RealD *p = OutRun->p_OUT[eSW_CO2Effects][pd];
 
 #if defined(RSOILWAT)
     get_outvalleader(
-        &sw->Model, pd, go->irow_OUT, go->nrow_OUT, go->tOffset, p
+        &sw->Model,
+        pd,
+        OutRun->irow_OUT,
+        OutDom->nrow_OUT,
+        sw->OutRun.tOffset,
```
dschlaep, line 218: Since we define `*OutRun`, we could do `OutRun->tOffset` instead of `sw->OutRun.tOffset`

src/SW_Output.c
dschlaep: Good suggestion, let's do that!

dschlaep (commented):
A few minor comments plus thinking about the suitable location for having code for the "construction" of SW_OUT_DOM

include/SW_datastructs.h
```diff
@@ -1328,13 +1314,67 @@ typedef struct {
...
 /* =================================================== */
-/*                 Comprehensive struct                */
+/*               Simulation Run Structs               */
 /* --------------------------------------------------- */
 
 typedef struct {
+
+    TimeInt tOffset; /* 1 or 0 means we're writing previous or current period */
+
+    char sw_outstr[MAX_LAYERS * OUTSTRLEN];
```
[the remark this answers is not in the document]
N1ckP3rsl3y: Wouldn't it work the same if we just use SW_OUTTEXT due to the definition of it being
`(defined(SOILWAT) || defined(STEPWAT)) && !defined(SWNETCDF)`?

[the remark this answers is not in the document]
dschlaep: Yes, indeed! `#id defined(SW_OUTTEXT)` is sufficient, thanks

dschlaep (approved):

Issue #425, opened 2024-08-01
nc-based SOILWAT2 runs slow down with increasing domain size


dschlaep:
The larger the simulation domain, the slower each simulation run becomes when using nc-output from SOILWAT2 v8.0.0. This is not the case with txt-based output. For instance:

| mode | nDimS | time_mean_s | time_min_s | time_max_s |
|------|-------|-------------|------------|------------|
| nc | 10 | 3.556 | 2.159 | 4.444 |
| nc | 10 | 3.532 | 2.187 | 4.375 |
| nc | 10 | 3.613 | 3.057 | 4.381 |
| nc | 20 | 5.29 | 3.298 | 6.449 |
| nc | 20 | 5.32 | 3.308 | 6.586 |
| nc | 30 | 6.62 | 4.605 | 7.43 |
| nc | 30 | 6.426 | 4.647 | 6.768 |
| text | 100 | 0.673 | 0.651 | 0.71 |
| text | 100 | 0.713 | 0.68 | 1.087 |
| text | 100 | 0.665 | 0.645 | 0.732 |
| text | 200 | 0.666 | 0.646 | 0.822 |
| text | 200 | 0.689 | 0.658 | 1.301 |
| text | 200 | 0.671 | 0.648 | 0.758 |

dschlaep:
Problem was that default chunking created chunks across simulation units -- this together with deflation caused that writing output of one run depended on total number of runs.

Proposed solution:
--> use custom chunking, e.g., `[1, 1, all time, all veg, all soil layers]` for xy and `[1, all time, all veg, all soil layers]` for site domains
--> new user input for deflate level: 0-9 where 0 means no deflation (input via Input_nc/desc_nc.in)

PR #428, opened 2024-08-07
Fix increasing write times with nc outputs

N1ckP3rsl3y:
- In the latest version of SOILWAT2 (v8.0.0), when the user creates a large domain, there is a noticeable increase in the runtime of individual simulations

- SOILWAT2 now makes use of
    - Custom nc chunking
    - A new user input - `deflateLevel`
    - An initial write to all output nc files while creating them
- See #425 for more information

### DIFF
--- a/src/SW_netCDF.c
+++ b/src/SW_netCDF.c
@@ -1337,7 +1337,14 @@ static void create_netCDF_var(
     if (!isnull(chunkSizes)) {
         if (nc_def_var_chunking(*ncFileID, *varID, NC_CHUNKED, chunkSizes) !=
             NC_NOERR) {
-            LogError(LogInfo, LOGERROR, "Just here.", varName);
+
+            LogError(
+                LogInfo,
+                LOGERROR,
+                "Could not chunk variable '%s' when creating it in "
+                "output netCDF.",
+                varName
+            );
             return; // Exit prematurely due to error
         }
     }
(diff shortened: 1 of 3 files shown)


src/SW_netCDF.c
```diff
@@ -1297,12 +1334,21 @@ static void create_netCDF_var(
         return; // Exit prematurely due to error
     }
 
+    if (!isnull(chunkSizes)) {
+        if (nc_def_var_chunking(*ncFileID, *varID, NC_CHUNKED, chunkSizes) !=
+            NC_NOERR) {
+            LogError(LogInfo, LOGERROR, "Just here.", varName);
```
dschlaep, line 1340: Could you please update the message on failed chunking to include a bit more information, e.g., "Failed to chunk variable '%s' in output netCDF."

src/SW_netCDF.c
```diff
@@ -1751,6 +1797,8 @@ static void fill_domain_netCDF_vals(
 @param[in] primCRSIsGeo Specifies if the current CRS type is geographic
 @param[in] domType Type of domain in which simulations are running
     (gridcell/sites)
+@param[asdf] deflateLevel Level of deflation that will be used for the created
```
dschlaep, line 1800: Could you please fix `@param[asdf] deflateLevel ...` in the documentation to `@param[in] deflateLevel ...`? Here and elsewhere including similar cases such as `@param[adf] deflateLevel ...` and `@param[d] deflateLevel ...`

dschlaep (commented):
Very nice! I've tested this for a domain of size 1,000,000 with and without deflation. It is now steadily making progress, excellent!

I made two minor comments about improving documentation and an error message. Please take a look at those and then we can finish this up. Thanks!

dschlaep (approved):

Issue #424, opened 2024-07-31
SW_NC_read_out_vars() doesn't close file connection


dschlaep:
It appears that SW_NC_read_out_vars() was missed when commit 6964cfb17805d33c530ac7bc757bf1d8fe483af0 "Use goto to close files in read files" was applied. We should fix that

PR #429, opened 2024-08-07
Close open files on error within `SW_netCDF.c`

N1ckP3rsl3y:
- In SOILWAT2 v8.0.0, there is a bug where the two functions - `SW_NC_write_output()` & `SW_NC_read_out_vars()` - upon error, the open file does not get closed

- The mentioned two functions now close files upon failure before leaving the function
- See #424 and #426


dschlaep (changes requested):
Looks great, thanks! All changes to file "SW_Control.c", however, are temporary debugging code that should not be here, please remove.

Issue #389, opened 2024-01-10
Capability to read inputs from netCDFs


dschlaep:
- Implement functionality to read inputs from netCDFs (if SWNETCDF)
- Don’t read daily weather from text files for `sw_template` if all weather is provided by netCDFs
- domainType vs. input types
    - a “sites” domain can read from input netCDFs that are “sites” or “xy”; if input netCDFs are of type “sites”, then they must match dimensions of domain
    - a “xy” domain requires “xy” input netCDFs
- Input netCDFs may be larger, smaller, only partially overlap in space and time with the domain
- We require that input netCDFs contain a set of CRS/XY-axis variables that matches up to such a set in the domain netCDF (fail if CRS not matching)
- Maybe: Pre-calculate index positions and store in a “index” netCDF (or use a user-provided file)
- As we loop over all simulation units in the domain, extract values for that suid/simulation run from each available input netCDFs
    - Retrieve index position from “index” netCDF
    - Read value(s)
    - Fail the run if no value (e.g., _FillValue)
    - Convert to SOILWAT2 units
    - Update `local_sw`
    - Run our checks on inputs (again) after values are updated from netCDFs

PR #432, opened 2024-08-26
Input netCDFs

N1ckP3rsl3y:
- SOILWAT2 currently has functionality to output simulated data through the netCDF (.nc) format (https://github.com/DrylandEcology/SOILWAT2/pull/390)
- Similarly, SOILWAT2 should allow the user to provide netCDF files with input data to use for the simulation which will replace the text (we use *.in) input file format
- Allow user to choose which input variables are provided by netCDFs while the others are input through text files
- See #389


src/SW_netCDF_Input.c
```diff
@@ -0,0 +1,3404 @@
...
+
+/**
+@brief Given a certain input key, make sure that specified input columns
+are the same throughout each active input variable
+
+@param[in] inputInfo Attribute information for all input variables
+@param[in] readInVars Specifies which variables are to be read-in as input
+@param[out] LogInfo Holds information on warnings and errors
+*/
+static void check_inputkey_columns(
```
dschlaep, line 405: `check_inputkey_columns()` is currently too strict for "ncVAxisName"

A non-NA value of ncVAxisName is required for SW2 variables with pattern “\<veg\>.*” (e.g., “\<veg\>.transp_coeff”, “\<veg\>.fCover”) but must be NA for all other SW2 variables even if part of the same inKey (e.g., “\<veg\>.fCover” has a ncVAxisName value while "bareGround.fCover" has NA)

src/SW_Output.c
```diff
@@ -38,13 +38,14 @@ See the \ref out_algo "output algorithm documentation" for details.
...
 #include "include/SW_Defines.h"     // for eSWC, OutPeriod, NVEGTYPES
 #include "include/SW_Files.h"       // for eOutput, eSite
-#include "include/SW_Site.h"        // for echo_inputs
-#include "include/SW_Times.h"       // for Today, Yesterday
-#include "include/SW_VegEstab.h"    // for echo_VegEstab
-#include "include/SW_VegProd.h"     // for echo_VegProd
-#include "include/Times.h"          // for Time_days_in_month, WKDAYS
-#include <stdio.h>                  // for snprintf, fprintf, printf
-#include <string.h>                 // for strcmp, memccpy, memset
+#include "include/SW_netCDF_Input.h" // for eSW_InDomain, vNCdom
```
dschlaep, line 41: Let's move `#include "include/SW_netCDF_Input.h"` to within `#if defined(SWNETCDF)` as elsewhere

src/SW_netCDF_Input.c
N1ckP3rsl3y: I currently have this change to reflect the following conditions:

* If the input variable is a general one (`<veg>.*`), we check to see that the V-axis is not 'NA'
    * If it is NA, fail
* Otherwise, if the input variable within the two keys soil and vegetation (the keys that hold the general variables) is not general, we make sure that the V-axis is NA
    * If the value is not NA, fail

I would like to confirm whether this is the preferred route to handle this column for these two input keys, or if there is a different preference when handling the V-axis for handling both general and non-general variables.

dschlaep: The check needs to cover all inkeys: fail if either the input variable does not contain the pattern (\<veg\>.*) or the V-axis name is not 'NA' but not both or none (i.e., exclusive or); otherwise pass

src/SW_netCDF_Input.c
```diff
@@ -0,0 +1,3404 @@
...
+@param[in] domFileID Domain netCDF file identifier
+@param[in] nDomainDims Number of dimensions the domain variable will have
+@param[in] primCRSIsGeo Specifies if the current CRS type is geographic
+@param[in] domType Type of domain in which simulations are running
+    (gridcell/sites)
+@param[in] deflateLevel Level of deflation that will be used for the created
+variable
+@param[out] LogInfo Holds information on warnings and errors
+*/
+static void fill_domain_netCDF_domain(
```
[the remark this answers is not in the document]
N1ckP3rsl3y: Did you want this to be the same for the `grid_mapping` attribute? More specifically, we have the value for a projected crs to be `"crs_projsc: x y crs_geogsc: <y name> <x name>"`. Did we want to keep the "x" and "y" in this?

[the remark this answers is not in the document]
N1ckP3rsl3y: Is it safe to presume that we will use these names rather than the ones that will be provided in the input spreadsheet and if so we would simply check that `geo_*AxisName` is the same as the X and Y axis names from said spreadsheet?

I would also like to confirm the ordering of these names regardng the coordinate attributes. We currently list the grid map as `"crs_projsc: x y crs_geogsc: lat lon"`. If I'm not mistaken, these coordinates are opposite of each other (i.e., not `x y lon lat` or `y x lat lon`). Is the current ordering correct?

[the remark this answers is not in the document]
dschlaep: For the first point: yes, we should be using the inputs from desc_nc.in for the domain because the spreadsheet does not provide all five; then check that what the spreadsheet provides is consistent with desc_nc.in (see section "check domain for consistent horizontal axes names" of the nc-input plan).

For the second point: the current order is correct (e.g., see example 5.13 in [CF-1.11](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.11)), i.e., the order is x-axis y-axis for projected crs and latitude longitude for geographic crs (in our case of wgs84)

src/SW_netCDF_Input.c
```diff
@@ -0,0 +1,3540 @@
...
+    int attNum;
+    int mustTestAttInd[] = {
+        SW_INNCVARNAME,
+        SW_INGRIDTYPE,
+        SW_INCRSNAME,
+        SW_INCRSEQUIV,
+        SW_INXAXIS,
+        SW_INYAXIS
+    };
+    const char *mustTestAttNames[] = {
```
dschlaep, line 295: Please update all code to handle the recent changes to the input.tsv including 
- remove: ncGridType, ncCRSEquivalency
- new: ncDomainType, ncSiteName, ncCRSGridMappingName

src/SW_netCDF_General.c
```diff
@@ -0,0 +1,1402 @@
...
+    Bool geoIsPrimCRS =
+        SW_Domain->OutDom.netCDFOutput.primary_crs_is_geographic;
+    char strAttVal[LARGE_VALUE];
+    double doubleAttVal;
+    const char *geoCRS = "crs_geogsc";
+    const char *projCRS = "crs_projsc";
+    Bool geoCRSExists = SW_NC_varExists(ncFileID, geoCRS);
+    Bool projCRSExists = SW_NC_varExists(ncFileID, projCRS);
+    const char *impliedDomType =
+        (SW_NC_dimExists("site", ncFileID)) ? "s" : "xy";
```
dschlaep, line 245: here and throughout, please use user-provided name for "site" dimension/variable (domain.in:siteName) instead of hard-coded "site"

src/SW_netCDF_Input.c
```diff
@@ -0,0 +1,3540 @@
...
+/**
+@brief Fill the variable "domain" with it's attributes
+
+@param[in] domainVarName User-specified domain variable name
+@param[in,out] domID Identifier of the "domain"
+    (or another user-specified name) variable
+@param[in] domDims Set dimensions for the domain variable
+@param[in] readinGeoYName User-provided geographical y-axis name
+@param[in] readinGeoXName User-provided geographical x-axis name
+@param[in] readinProgYName User-provided projected y-axis name
```
dschlaep, line 852: Please correct `readinProgXName` and `readinProgYName` to `readinProjXName` and `readinProjYName` -- the "proj" refers to "projected"

src/SW_netCDF_Input.c
```diff
@@ -0,0 +1,3540 @@
...
+        return; /* Exit function prematurely due to error */
+    }
+
+    SW_NCIN_alloc_input_var_info(SW_netCDFIn, LogInfo);
+    if (LogInfo->stopRun) {
+        return; /* Exit function prematurely due to error */
+    }
+
+    /* Ignore the first row (column names) */
+    GetALine(f, inbuf, MAX_FILENAMESIZE);
```
dschlaep, line 2934: Our current code only works for the hard-coded order of columns in input.tsv. If the user changes the column order, the code produces error messages that are not helpful, e.g., "Start year of weather input file for 'tasmax' is <= 0."

I suggest that `SW_NCIN_read_input_vars()` checks that the column names are in the exact order that the code expects and fails otherwise. Similarly, `SW_NCOUT_read_out_vars()` should also check the column names.

dschlaep (commented):
Great progress and thanks for addressing comments! I marked some of my comments as resolved, and I left others open because I currently cannot run the code and check behavior. I added a few more comments.

Additionally, both the test program and the simulation program fail and make checking/review difficult:

* The test program in nc-mode `make CPPFLAGS=-DSWNC test_run` fails with
```
error: assigning to 'char *' from incompatible type 'void *'
            tempCoordPtr = sw_memccpy(writePtr, " ", '\0', writeSize);
```

* The simulation program in nc-mode (`make CPPFLAGS=-DSWNC bin_run`) fails with a variety of errors likely because of mismatches between the different versions of the input tsv and the code. Please update the code to handle our current version of the input tsv. Thanks!

src/SW_netCDF_Input.c
```diff
@@ -0,0 +1,6230 @@
...
+            domCoordVarNames[index],
+            *domCoordArrs[index],
+            LogInfo
+        );
+        if (LogInfo->stopRun) {
+            return; /* Exit prematurely due to error */
+        }
+
+        for (val = 0; val < *domCoordSizes[index]; val++) {
+            /* Check latitude */
```
dschlaep, line 2313: read_domain_coordinates() [as commit d09004e0d1e818b19eb5cdd96fb942a90c0a96ae] loops through all coordinate variables and checks variables associated with the Y-axis dimension to be within [-90, 90] and variables associated with the X-axis dimension to be within [-180, 180].

There are two problems with the current version
* The limits [-90,90] and [-180,180] are only valid for geographic but not projected coordinate variables --> skip these checks for projected coordinate variables.
* There are two conventions for longitude [-180,180] and [0,360], and more generally, longitude can "wrap" around (i.e., `(180 + lon % 360) - 180`) --> don't check longitude values.

Thus, the only remaining check would be for the geographic Y-axis variable latitude. Not sure the for-loop is warranted. 

Thanks!

src/SW_netCDF_Input.c
```diff
@@ -0,0 +1,8443 @@
...
+         "m s-1",
+         "m s-1",
+         "m s-1",
+         "%",
+         "%",
+         "%",
+         "%",
+         "degC",
+         "kPa",
+         "m s-1"},
```
dschlaep, line 98: The hard-coded units for shortWaveRad should be "NA" instead of current "m s-1" (see tsv nc-inputs) -- and the code should skip the check if the units are matching (the same as for the units of the tsv nc-inputs swrcp[*]). This is because different datasets use different approaches for solar radiation (flux density [W m-2] vs. irradiation [MJ m-2]) which we handle via `desc_rsds`.

src/SW_Site.c
```diff
@@ -181,68 +181,74 @@ static Bool SW_check_soil_properties(
...
+        fval = SW_Site->soils.fractionWeight_om[layerno];
         errtype = Str_Dup("organic matter content", LogInfo);
 
-    } else if (LT(SW_Site->evap_coeff[layerno], 0.) ||
-               GT(SW_Site->evap_coeff[layerno], 1.)) {
+    } else if (LT(SW_Site->soils.evap_coeff[layerno], 0.) ||
+               GT(SW_Site->soils.evap_coeff[layerno], 1.)) {
         res = swFALSE;
-        fval = SW_Site->evap_coeff[layerno];
+        fval = SW_Site->soils.fractionWeight_om[layerno];
```
dschlaep, line 236: SW_check_soil_properties(): this looks like an error during merge conflict resolution
```
    } else if (LT(SW_Site->soils.evap_coeff[layerno], 0.) ||
               GT(SW_Site->soils.evap_coeff[layerno], 1.)) {
        res = swFALSE;
        fval = SW_Site->soils.fractionWeight_om[layerno];
        errtype = Str_Dup("organic matter content", LogInfo);
```
and should be removed: fractionWeight_om is checked on lines 227-231 and evap_coeff is checked on lines 239-243 (see also https://github.com/DrylandEcology/SOILWAT2/commit/8f895992f208793d33d92746c3eca9382febec9f#diff-f9371e34e93e7bab2198e93bdaccdbd4186f5127f01d95aa44da7ea9d26db0e5)

src/SW_Site.c
```diff
@@ -2537,18 +2556,18 @@ void SW_SIT_init_run(
...
-            SW_Site->fractionWeight_om[s],
+            SW_Site->soils.fractionWeight_om[s],
             SW_Site->depthSapric,
-            (s > 0) ? SW_Site->depths[s - 1] : 0,
-            SW_Site->depths[s]
+            (s > 0) ? SW_Site->soils.depths[s - 1] : 0,
+            SW_Site->soils.depths[s]
         );
 
         /* Check parameters of bulk soil SWRC */
```
dschlaep, line 2568: The `/* Check parameters of bulk soil SWRC */` should use `SW_Site->swrcp[s]` and not `SW_Site->swrcpMineralSoil[s]` -- see https://github.com/DrylandEcology/SOILWAT2/commit/f254f40d04b7200eec5229e00fcd6e34d0c5f8e1#diff-f9371e34e93e7bab2198e93bdaccdbd4186f5127f01d95aa44da7ea9d26db0e5

dschlaep (changes requested):
Lots of excellent work, thanks! Some requests and suggestions

src/SW_netCDF_Input.c
```diff
@@ -0,0 +1,8655 @@
...
+
+    const char *const acceptableCals[] = {
+        /* 365 - 366 days */
+        "standard",
+        "gregorian",
+        "proleptic_gregorian",
+
+        /* Always has 366 days and will discard values */
+        "all_leap",
+        "alleap",
```
dschlaep, line 3028: determine_valid_cal(): please change "alleap" to "allleap"

src/SW_netCDF_Input.c
```diff
@@ -0,0 +1,8903 @@
...
+            ncSuid,
+            (void *) &domStatus,
+            "unsigned integer",
+            LogInfo
+        );
+        if (LogInfo->stopRun) {
+            goto freeMem; // Exit function prematurely due to error
+        }
+
+        vals[suid] = (domStatus == NC_FILL_UINT) ? NC_FILL_BYTE : PRGRSS_READY;
```
dschlaep, line 2121: fill_prog_netCDF_vals(): the check `(domStatus == NC_FILL_UINT)` does not work correctly for a "domain.nc" that was produced (e.g., outside SOILWAT2) with a "domain" variable that is not unsigned int.

--> can we query the type of "domain" variable for the call SW_NC_get_single_val() and obtain the _FillValue attribute of the "domain" variable (if available) to check  `(domStatus == fillValue)`

dschlaep (commented):
Just about there!

dschlaep (approved):

src/SW_Output.c
dschlaep, line 1: SW_OUTDOM_construct(): 
```
#if defined(RSOILWAT) || defined(SWNETCDF)
            OutDom->pfunc_mem[k] =
                (void (*)(OutPeriod, SW_RUN *, SW_OUT_DOM *)) get_none_outarray;
```
fails because `get_none_outarray()` is now defined with additional argument `LOG_INFO *`

Issue #397, opened 2024-04-30
Organic soil content: influence on soil water retention


dschlaep:
- New soil property: soil organic content (default input: 0)

- Approach that linearly combines mineral and peat properties based on fraction of soil organic content (see https://doi.org/10.1007/s00382-007-0278-1)

Literature
- Cambpell’s SWRC: https://doi.org/10.1007/s00382-007-0278-1
- van Genuchten’s SWRC: https://doi.org/10.1080/07055900.2000.9649643

PR #434, opened 2024-10-21
Feature soil organic content

dschlaep:
See https://github.com/DrylandEcology/SOILWAT2/issues/397

* SOILWAT2 now represent the influence of soil organic matter on the soil water retention curve and the saturated hydraulic conductivity parameter
* Saturated percolation is now limited. The upper bound is a function based on the saturated hydraulic conductivity parameter (which includes effects of organic matter), frozen soils, and a user-specified `"permeability"` factor.

* New input via `"siteparam.in"` to specify the depth at which characteristics of organic matter have completely switched from fibric to sapric peat
* New input via `"soils.in"` to provide the proportion of soil organic matter to bulk soil by weight for each soil layer.
* New input via `"swrc_params*.in"` to provide parameter values of the soil water retention curve representing fibric and sapric peat.


tests/gtests/test_SW_Flow_Lib.cc
```diff
@@ -165,17 +165,19 @@ TEST(SWFlowTest, SWFlowSaturatedPercolation) {
...
-    double standingWater;
+    double standingWater = 0.;
     double drainout;
+    const double swc0init = 0.8;
 
     // ***** Tests when nlyrs = 1 ***** //
     ///  provide inputs
     int nlyrs = 1;
-    double swc[1] = {0.8};
+    double swc[1];
```
N1ckP3rsl3y, line 175: On my end, the command `CXX=g++ make clean test_run` results in this test to fail. If I initialize the variable `swc` to `{0}`, the test passes. Is this reproducible?

N1ckP3rsl3y (changes requested):

tests/gtests/test_SW_Flow_Lib.cc
dschlaep: thanks! Commit 861950ef074f6254079a1b1f582376efb876326d aims to address this. This didn't show up for me with g++ v14.2.0

N1ckP3rsl3y: That's interesting, I was using g++ v12.4.0, however, it is now gone!

N1ckP3rsl3y (approved):

Issue #440, opened 2025-01-03
Temperature at surface and in soil explodes under huge biomass


dschlaep:
Demonstration with R code using rSOILWAT2 v6.2.0

```
  tmp <- rSOILWAT2::sw_exampleData
  tmp@prod@MonthlyVeg[["Shrubs"]][, "Live_pct"] <- 50
  tmp@prod@MonthlyVeg[["Shrubs"]][, "Biomass"] <- 10000
  tmp@prod@MonthlyVeg[["Shrubs"]][, "Biomass"] <- 50000
  x <- rSOILWAT2::sw_exec(tmp)
  graphics::matplot(x@TEMP@Day[1:365, 6:8], type = "l", col = c("orange", "skyblue", "black"))
  graphics::matplot(x@SOILTEMP@Day[1:365, -(1:2)], type = "l")
```

PR #445, opened 2025-01-24
Bugfix440 soil temperature

dschlaep:
Fi estimation of surface temperature (#440):

- Simulation output is slightly changed
- Biomass effects are now capped at a value at which cooling and heating effects on minimum and maximum surface temperature result, across average conditions, in no change for mean surface temperature
- Effects of maximum air temperature on maximum surface temperature are limited to air temperatures above freezing. Minimum and maximum surface temperature are now set to their average (with a warning) if the initial estimate of minimum surface temperature is larger than the initial estimate of maximum surface temperature
- Average surface temperature is now (by default) estimated from minimum and maximum surface temperature (instead of independently as previously), but the old method can be selected.
- New input: methodSurfaceTemperature


src/SW_Flow_lib.c
```diff
@@ -3277,6 +3277,21 @@ void soil_temperature(
     for (i = 0; i < nlyrs; i++) {
         maxTempSoil[i] = meanTempSoil[i] + (temperatureRange[i] / 2);
         minTempSoil[i] = meanTempSoil[i] - (temperatureRange[i] / 2);
+
+        if (minTempSoil[i] < -100 || maxTempSoil[i] > 100) {
```
N1ckP3rsl3y, line 3320: I suggest we use `LT()` and `GT()` in instances all instances except for the ones checking for a positive value, since this does not rely on floating point precision.

N1ckP3rsl3y (approved):

PR #455, opened 2025-02-25
Test Cygwin with Structure Reconfiguration

N1ckP3rsl3y:
* Attempt to find the reason Cygwin is resulting in a segmentation fault when reconfiguring structs


dschlaep:
See unresolved issue https://github.com/DrylandEcology/SOILWAT2/issues/478

PR #495, opened 2025-12-11
Simplify output algorithm

### LINKED ISSUE #494: Establishment output is incorrectly set for previous period
Within the example weekly output results, the column(s) ESTBL_* can result in incorrect values for weekly values. This is a result of how output is being done. When output is stored through `get_*()`functions, usually output structs are used, but with establishment, it's `SW_VEGESTAB_SIM` that is used. This means since output is stored after a period is completed, `SW_VEGESTAB_SIM` contains the current period data rather than data for the previous. This then results in the possibility for reporting a value one output period earlier than it should.

N1ckP3rsl3y:
Simplify the output algorithm to remove the need to access the "previous" output period and redundancy in calls to output information

- Remove use/mention of `tOffset`, `bFlush_output` and `SW_OUT_flush()` which add complexity to the output algorithm
- Update the output algorithm to collect/average values at the end of the each period rather than at the beginning of the next
- Update the documentation on the output algorithm to remove outdated information and to be up-to-date as of this PR

- Fix an unknown bug where establishment values had the chance to be reported one period before they should be (see commit and issue: 657be4e979869dbd2294c2dde4dd344972afaf61 & #494)
- close #494


src/SW_Output.c
```diff
@@ -680,14 +675,14 @@ static void average_for(
 
         switch (pd) {
         case eSW_Week:
-            curr_pd = (sw->ModelSim.week + 1) - tOffset;
-            div = (bFlush_output) ? sw->ModelSim.lastdoy % WKDAYS : WKDAYS;
+            curr_pd = sw->ModelSim.week + 1;
+            div = (doy == lastDoy) ? lastDoy % WKDAYS : WKDAYS;
```
dschlaep: In average_for(), the weekly `div` is calculated as `lastDoy % WKDAYS` if doy is the lastDoy. This seems to be wrong (this has been in the code for a very long time seemingly!): if lastDoy is the last day in a week then, then div = 0 (instead of 7 and it results in division by zero). Shouldn't this be `(lastDoy - 1) % WKDAYS + 1`?

For instance, setting `EndDoy  350` in "Inputs/domain.in" produces "Inf" for the week 50 in "Output/sw2_weekly.csv"

However, this branch is now producing monthly output only for a month if the simulation completed the entire month (`SW_MDL_new_day()`). I like that better than weekly output which is currently output even if a week is incomplete.

dschlaep (approved):
This is a great PR! It's really a neat simplification. I added one comment about problems with output of weeks that are partial or when the simulation ends before the end of the calendar year. I addressed my comment directly with the commit 443f688591c2640f5054d4775a1d2a92f4701ac8 and approve of this PR.

PR #499, opened 2025-12-12
Fix Markov generator

### LINKED ISSUE #497: Incorrect deep copying of markov information
- When deep copying the template SW_RUN in `SW_RUN_deepCopy()`, SW_MARKOV_INPUTS is a part of this
- However, when the weather generator method is 2 (only copied with this value), the dynamically allocated part(s) of the struct are not allocated before copying so the following sequence occurs
   - Deep copy into the template instance of the program (the same array)
   - Complete a site simulation
   - Deallocate markov information (within the template SW_RUN)
   - Exit the program and attempt to deallocate markov, resulting in the following error locally

```
SOILWAT2(18791,0x7ff848eccfc0) malloc: double free for ptr 0x7fe815060e00
SOILWAT2(18791,0x7ff848eccfc0) malloc: *** set a breakpoint in malloc_error_break to debug
./tools/run_bin.sh: line 62: 18791 Abort trap: 6           bin/SOILWAT2 -d ./tests/example -f files.in -r
make: *** [bin_run] Error 134
```

### LINKED ISSUE #498: Weather generator method = 2 produces an error
- When running the example data with weather generator method = 2 in SWNETCDF mode, it results in the following error reported in the log file
```
(suid = 1) ERROR: Missing weather data (day 1980 - 1) during simulation: Tavg=999.00, ppt=999.00, wind=1.30, rH=61.00, vp=999.00, rsds=999.00 / cloud=66.48
```

- This is because when reading netCDF inputs, we do not finalize the weather if the weather generator is turned off entirely

### LINKED ISSUE #500: NetCDF test runs do not create reference in `SWMPI` mode
When running `./tools/ncTestRuns.sh cleanReference all --mode=mpi`, the error message of `Failed to locate reference output ...` is produced. This is caused by a couple reasons

1) The call to the r-script `Rscript__ncTestRuns_00_createReferenceRun.R` through `tools/ncTestRuns.sh` provides an "ntasks" value of 2, rather than 1
2) `Rscript__ncTestRuns_00_createReferenceRun.R` does not look for the base logfile name in SWMPI mode ("rank_0_logfile.log"), rather it looks for "logfile.log", returning a successful run status shortly resulting the error provided above

Previous work
1) The commit https://github.com/DrylandEcology/SOILWAT2/commit/b3d079ccf1fa83947bc0e4615da3beb0e34931dd worked to update the code, but did not cover the reference run information

N1ckP3rsl3y:
A few bugs have been found that deal with the markov weather generator which need to be fixed

- When the weather generator method is 2 (markov), dynamic memory is not allocated within `SW_RUN_deepCopy()` resulting in a possible segmentation fault or attempting to free a pointer that's already been freed (see #497)
- In SWNETCDF mode, we do not finalize the weather if weather is not read in (see #498)

### DIFF
--- a/include/SW_Weather.h
+++ b/include/SW_Weather.h
@@ -46,6 +46,16 @@
 extern "C" {
 #endif
 
+/* Weather generation methods */
+
+/** No weather generation, only historical data */
+static const unsigned int wgHist = 0;
+
+/** Weather generation method LOCF, see generateMissingWeather() */
+static const unsigned int wgLOCF = 1;
+
+/** Markov weather generator method, see generateMissingWeather() */
+static const unsigned int wgMKV = 2;
 
 /*  all temps are in degrees C, all precip is in cm */
 /*  in fact, all water variables are in cm throughout
--- a/src/SW_netCDF_Input.c
+++ b/src/SW_netCDF_Input.c
@@ -8579,25 +8579,6 @@ void SW_NCIN_read_inputs(
         checkReturn(mainLogInfo->stopRun);
     }
 
-    if (runSims && readWeather) {
-        for (input = 0; input < numInputs; input++) {
-            SW_WTH_finalize_all_weather(
-                &sw->MarkovIn,
-                &sw->WeatherIn,
-                inputs[input].weathRunAllHist,
-                sw->ModelSim.cum_monthdays,
-                sw->ModelSim.days_in_month,
-                domSuids[input],
-                SW_Domain->netCDFInput.siteDoms[eSW_InDomain],
-                mainLogInfo
-            );
-            if (mainLogInfo->stopRun) {
-                break;
-            }
-        }
-        checkReturn(mainLogInfo->stopRun);
-    }
-
     if (runSims && readVeg) {
         read_veg_inputs(
             SW_Domain,
--- a/src/SW_netCDF_Input.c
+++ b/src/SW_netCDF_Input.c
@@ -8577,7 +8577,9 @@ void SW_NCIN_read_inputs(
             mainLogInfo
         );
         checkReturn(mainLogInfo->stopRun);
+    }
 
+    if (runSims && readWeather) {
         for (input = 0; input < numInputs; input++) {
             SW_WTH_finalize_all_weather(
                 &sw->MarkovIn,
--- a/tests/ncTestRuns/scripts/Rscript__ncTestRuns_01_createTestRuns.R
+++ b/tests/ncTestRuns/scripts/Rscript__ncTestRuns_01_createTestRuns.R
@@ -403,16 +403,23 @@ pb <- utils::txtProgressBar(max = nrow(listTestRuns), style = 3L)
 for (k0 in seq_len(nrow(listTestRuns))) {
 
   #------ . ------
-  # Skip a testRun if no external weather data
-  if (isFALSE(hasWeather[[listTestRuns[k0, "inWeather"]]])) {
+  useWeatherGenerator <- identical(listTestRuns[k0, "inWeather"], "wGen")
+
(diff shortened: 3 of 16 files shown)


include/SW_Weather.h
```diff
@@ -46,6 +46,16 @@
...
+/* Weather generation methods */
+
+/** No weather generation, only historical data */
+static const unsigned int wgHist = 0;
+
+/** Weather generation method LOCF, see generateMissingWeather() */
+static const unsigned int wgLOCF = 1;
+
+/** Markov weather generator method, see generateMissingWeather() */
+static const unsigned int wgMKV = 2;
```
dschlaep, line 58: - SW_MKV_setup() has two instances with `Weather_genWeathMethod == 2` that should now be `Weather_genWeathMethod == wgMKV`
- generateMissingWeather() has `method == 0` that should now be `method == wgHist`; it also has `method > 2` that should probably be `method != wgHist && method != wgLOCF && method != wgMKV`

src/SW_netCDF_Input.c
```diff
@@ -8577,7 +8577,9 @@ void SW_NCIN_read_inputs(
             mainLogInfo
         );
         checkReturn(mainLogInfo->stopRun);
+    }
 
+    if (runSims && readWeather) {
         for (input = 0; input < numInputs; input++) {
             SW_WTH_finalize_all_weather(
                 &sw->MarkovIn,
```
dschlaep: Doesn't a SW_NCIN_read_inputs() with `for (input ...) SW_WTH_finalize_all_weather(&sw->MarkovIn, ...) ` causes MarkovIn.markov_rng and maybe(?) markov_rng.ppt_events to be shared among the `numInputs`?

It seems we need a new struct SW_MARKOV_RUN_INPUTS that contains markov_rng and ppt_events which then becomes part of SW_RUN_INPUTS (and/or add SW_MARKOV_INPUTS to SW_RUN_INPUTS)? Then, this loop would become `for (input ...) SW_WTH_finalize_all_weather(inputs[input].MarkovRunIn, ...)`

dschlaep (approved):

tests/ncTestRuns/scripts/Rscript__ncTestRuns_01_createTestRuns.R
```diff
@@ -403,16 +403,23 @@ pb <- utils::txtProgressBar(max = nrow(listTestRuns), style = 3L)
 for (k0 in seq_len(nrow(listTestRuns))) {
 
   #------ . ------
-  # Skip a testRun if no external weather data
-  if (isFALSE(hasWeather[[listTestRuns[k0, "inWeather"]]])) {
+  useWeatherGenerator <- identical(listTestRuns[k0, "inWeather"], "wGen")
```
N1ckP3rsl3y, line 408: When I run the command `./tools/check_ncTestRuns.sh cleanReference clean all`, upon the rest of the output, I get the following seemingly duplicated lines

```
   testRun-38     success       ok     values: ok            ok
   testRun-38     success       ok     values: ok            ok
   testRun-40     success       ok     values: ok            ok
   testRun-40     success       ok     values: ok            ok
```
where each line doesn't reuse an existing folder (i.e., two unique folders are created for test 38).

When I run `./tools/check_ncTestRuns.sh cleanReference clean all -t 39` I get the following error

```
Create ncTestRuns ...
Error in utils::txtProgressBar(max = nrow(listTestRuns), style = 3L) : 
  must have 'max' > 'min'
Execution halted

Execute and check ncTestRuns ...
Error: any(lengths(fnames_ref) > 0L) is not TRUE
Execution halted
```

Was this an intended modification?

PR #505, opened 2025-12-19
Allow last day of last year in netCDF files

### LINKED ISSUE #185: Remove `start` and `end` from `outsetup.in`
With the overhaul of the output, i.e., from individual files to `csv`-files that collect output from all output types, the code assumes that every output variable uses the same `start` and `end` day of year.

Plus, the code for weekly and monthly output never correctly checked whether the corresponding day was within `start` and `end` day of year, e.g., 
```
curr_pd = (SW_Model.month + 1) - tOffset;
if (SW_Output[k].myobj != otyp || curr_pd < SW_Output[k].first || curr_pd > SW_Output[k].last)
    continue;
```
This doesn't work correctly because `curr_pd` in this case is between 1 and 12 (month), whereas `SW_Output[k].first` and `SW_Output[k].last` are between 1 and 366 (doy)!

--> eliminate `SW_Output[k].first` and `SW_Output[k].last` and have code produce output for every day of each year (as is the current defaults for `outsetup.in` of `start = 1` and `end = end`), i.e., use `SW_Model.firstdoy` and `SW_Model.lastdoy` instead.

### LINKED ISSUE #503: Capability to end simulation on any day of year

N1ckP3rsl3y:
Fixes #503

- Adjust output netCDF file temporal sizes based on how many days are in the last year
- No longer produce yearly data if it is partially completed
- Fix the first/last day of the first/last year of the simulation by properly determining when the simulation is in spinup mode


dschlaep (commented):
Could you please take another look -- there seem to be a few problems.

Providing EndDoy works in txt-mode with `make bin_run` -- it doesn't work with `make bin_debug_severe` which appears to ignore the value of EndDoy.

In nc-mode, the daily output stops one day too soon (e.g., output ends on Jan 30 if user provides EndDoy=31). Also, the values of all output periods appear to be garbled up. If I set EndDoy < 31, then I get a segmentation fault with `make bin_run`. I get a buffer overflow in get_temp_mem() with `make bin_debug_severe` for any value of EndDoy < 365.

dschlaep (commented):
Great, thanks! The commits have fixed the issues that I observed. I came across a new problem: Setting EndDoy to an early end date, works in txt-mode, but nc-mode produces output for incomplete weekly and monthly output for some but not all values of EndDoy.

For example, EndDoy = 9 (Jan-09) stops weekly and monthly output one time step too late, while daily and yearly are correct. Monthly output ends with 2010-01-15 (2009-12-31, 2010-01-31) and weekly output ends with 2010-01-10 (2010-01-07, 2010-01-14) -- both after Jan-09.

EndDoy = 91 (Apr-01) produce correct last time steps.
EndDoy = 90 (Mar-31) is incorrect for weekly output with 2010-03-28 (2010-03-25, 2010-04-01)
EndDoy = 89 (Mar-30) is incorrect for weekly output with 2010-03-28 (2010-03-25, 2010-04-01)
etc.

R code that I used to calculate end dates
```
path <- file.path("tests", "example", "Output")

fns <- list.files(path, pattern = "^AET_2000-2010_")

for (fn in fns) {
  cat("Last time step of", fn)
  
  xnc <- RNetCDF::open.nc(file.path(path, fn))
  x <- RNetCDF::read.nc(xnc)
  timeUnits <- RNetCDF::att.get.nc(xnc, "time", attribute = "units")
  nTime <- length(x[["time"]])
  
  timeValue <- RNetCDF::utcal.nc(timeUnits, x[["time"]][[nTime]], type = "s") |>
    as.Date() |>
    as.character()
  timeBounds <- RNetCDF::utcal.nc(timeUnits, x[["time_bnds"]][, nTime], type = "s") |>
    as.Date() |>
    as.character()
  
  cat(": ", timeValue, " (", toString(timeBounds), ")", sep = "", fill = TRUE)
}
```

dschlaep (approved):

PR #525, opened 2026-04-09
Time-before-space to space-before-time

### LINKED ISSUE #469: mpi-based SOILWAT2: segfault because N_SUID_ASSIGN is too large
Compiling mpi-based SOILWAT2 with a large value of N_SUID_ASSIGN can lead to segfaults. Depending on exact project setup and compile configurations values of 150-200 for N_SUID_ASSIGN have often caused segfaults.

### LINKED ISSUE #470: mpi-based SOILWAT2: slow performance/stalling with multiple nodes
Throughput of mpi-based SOILWAT2 drops or the program stalls if run on multiple nodes.

### LINKED ISSUE #485: Convert nc/mpi-mode SOILWAT2 to space-before-time

### LINKED ISSUE #502: Restart simulation from cache file
For nc-based, space-before-time SOILWAT2, store current state of simulation in a netCDF cache file. Then, restart from the cache file and continue the simulation as if the simulation never was stopped

N1ckP3rsl3y:
closes #485
addresses #469 
closes #470 

- The current approach to run simulations is to run one simulation site after another
    - This is not the approach the majority of other similar simulation softwares take (space-before-time)
- Switch to space-before-time approach, where all active simulation sites simulate one day at a time
- New feature/requirement is to cache any intermediate values of a simulation run if interrupted in the middle of it (cached_state.nc)
- Now calculate spatial/temporal netCDF chunking sizes that will best fit the number of created processes to have minimal overlap when reading/writing values


NEWS.md
```diff
@@ -688,7 +688,7 @@
       This function is based on `rSOILWAT2::estimate_PotNatVeg_composition()`
       which was previously coded in R.
     * New `estimateVegetationFromClimate()`
-      (which is called by `SW_VPD_init_run()`) uses `veg_method` to determine
+      (which is called by `SW_VPD_init_run_calc)`) uses `veg_method` to determine
```
dschlaep: Please undo the renamed function name in NEWS -- back for that old version, `SW_VPD_init_run()` was the correct name, don't change that

dschlaep (approved):

PR #528, opened 2026-07-09
Feature output packing

### LINKED ISSUE #527: Write netCDF output optionally as packed variables
Add capability to write output as packed variables if user requests packing for a variable.

The ncoutput.tsv gains three new columns

- output_type with possible values: NC_DOUBLE (default), NC_SHORT, or NC_INT -- if short or int (netCDF-types), then the variable will be packed
- scale_factor (default 1), used as attribute if variable is packed (cannot be 0)
- add_offset (default 0), used as attribute if variable is packed

The attributes scale_factor and add_offset are double (the unpacked variable type).
The packed values are calculated as: packed = round((original - add_offset) / scale_factor) # round to nearest integer, nearbyint()?
The _FillValue attribute is type "output_type" with value -32768 = -2^(16-1) for NC_SHORT and -2^(32-1) for NC_INT.

The user can specify for each output variable separately if a variable should be packed or not and each variable can have different scale_factor and add_offset values.

Comment for user:
- scale_factor = (max - min) / (2^bits - 1) where NC_SHORT has 16 and NC_INT 32 bits
- add_offset = (max + min) / 2.

N1ckP3rsl3y:
closes https://github.com/DrylandEcology/SOILWAT2/issues/527

- One problem with larger simulation domains is they create a vast amount of data, due to our outputs being type `double`
- A way to fix this is to pack the output values into smaller data types like `short` or `integer`
- This could decrease the total output size by at least half
    - NC_DOUBLE = 64 bits
    - NC_INT = 32 bits
    - NC_SHORT = 16 bits
 
- Available output types: double, integer and short


src/SW_netCDF_Output.c
```diff
@@ -1459,10 +1485,198 @@ static void create_output_file(
...
+
+@param[in] type Target type to convert to (short or integer)
+@param[in] nVals Number of values to convert
+@param[in] scale_factor Scale factor to use for conversion
+@param[in] add_offset Add offset to use for conversion
+@param[in] doubleVals Array of type double holding the values to convert
+@param[out] shortVals Array of type short to hold the converted values
+@param[out] intVals Array of type integer to hold the converted values
+*/
+static void pack_output_values(
```
dschlaep, line 1611: The _FillValue for packed short should be -32768 = -2^(16-1) and not NC_FILL_SHORT = -32767 (see issue #527 and https://docs.unidata.ucar.edu/nug/current/best_practices.html#bp_Packed-Data-Values "In either the signed or unsigned case, an alternate formula may be used for the add_offset and scale_factor packing parameters that reserves a packed value for a special value, such as an indicator of missing data." etc)
And the same for integer packed data.

src/SW_netCDF_Output.c
```diff
@@ -1459,10 +1485,198 @@ static void create_output_file(
...
+    int *intVals
+) {
+    size_t index;
+    double valToPack;
+
+    if (Str_CompareI(type, (char *) "short") == 0) {
+        for (index = 0; index < nVals; index++) {
+            if (!EQ(doubleVals[index], NC_FILL_DOUBLE)) {
+                valToPack = (doubleVals[index] - add_offset) / scale_factor;
+                shortVals[index] = (short) nearbyint(valToPack);
```
dschlaep: pack_output_values() currently does not guard against over-/underflow during packing.

A too small scale_factor as provided by the user leads with the coercion to short (or int) to incorrect values. For instance when packing to short, doubleVals = 0.15, add_offset = 0.5, scale_factor = 1e-5 -> valToPack = -35000 which is smaller than can be represented by a short. This likely ends up as 30536 which when unpacked becomes incorrect 30536 * 1e-5 + 0.5 = 0.80536 != 0.15.

These cases should be set to _FillValue and a warning issued. The same applies for packing to integer.

dschlaep (approved):

